PR #7, opened 2021-04-09
Fix linter issues

vmartinv:
I think the inconsistencies between isort and black are solved with the current version, at least from what I could test.
Also, I followed this [guide](https://github.com/github/super-linter/blob/master/docs/run-linter-locally.md) in order to run all the linters locally, it's a bit slow but good as a final check.


KapJI:
Thanks, I actually think we should get rid of super linter and setup linters manually. I'll probably try to do that.

PR #8, opened 2021-04-09
put parsers in a separate file

vmartinv:
This moves all the parsers into `parsers.py`. This way `calc.py` is smaller and the parsing logic is isolated from the calculations.
This is part of a series of changes that will add (in this order):
- More verbose exception handling (I had a hard time debugging without it)
- Trading212 support (only needs a parser, the rest of the code stays the same)
- Removal of global variables (this is to make testing possible)
- Unit tests (because I want to be sure that my taxes are calculated right)

Let me know what you think about these too!

### DIFF
--- a/calc.py
+++ b/calc.py
@@ -1,6 +1,5 @@
 #!/usr/bin/env python3
 
-import csv
 import datetime
 import decimal
 import sys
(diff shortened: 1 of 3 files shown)


KapJI:
Looks good, thanks!

Getting rid of global variables and adding argparse would be great. I think we can even build a proper package and publish it on PyPI. But I'm not sure what to do with input files like `initial_prices.csv` or `gbp_usd_history.csv`. Probably we should put gbp/usd prices in the package with the option to redefine it.

PR #9, opened 2021-04-11
update 2020/21 prices

vmartinv:
These are the prices missing for the last financial year, I got them from the HMRC website and from Schwab


calc.py
```diff
@@ -24,10 +24,10 @@
 )
 
 # First year of tax year
-tax_year = 2019
-# Allowance is £12000 for 2019/20
+tax_year = 2020
+# Allowance for 2020/21
 # https://www.gov.uk/guidance/capital-gains-tax-rates-and-allowances#tax-free-allowances-for-capital-gains-tax
-capital_gain_allowance = 12000
+capital_gain_allowance = 12300
```
KapJI: Let's create a map for this with year as a key and include allowances for different years.

vmartinv: Added suggestions

calc.py
```diff
@@ -34,10 +34,19 @@
...
 
 # First year of tax year
-tax_year = 2019
-# Allowance is £12000 for 2019/20
+tax_year = 2020
+# Allowances
 # https://www.gov.uk/guidance/capital-gains-tax-rates-and-allowances#tax-free-allowances-for-capital-gains-tax
-capital_gain_allowance = 12000
+capital_gain_allowances = {
+    2013: 12000,
```
KapJI: I think they changed rules here, let's drop this one.
```suggestion
```

KapJI:
Thanks!

PR #10, opened 2021-04-11
Use more verbose exceptions

vmartinv:
This adds more easy to understand errors. I didn't change all the asserts, just the ones that happened more often when I was debugging.


KapJI:
Looks great, thanks!

PR #11, opened 2021-04-11
make calc.py and model.py schwab agnostic

vmartinv:
This moves all the schwab specific logic to `schwab.py`. Also allows to track balances in different currencies during phase 1, this is needed because trading212 UK uses GBP as base currency while Schwab uses USD.
Final step before adding trading212 support.

Planned next diffs:
- remove global variables
- unit tests
- argparse


schwab.py
```diff
@@ -0,0 +1,92 @@
+import csv
+import datetime
+from decimal import Decimal
+from typing import List
+
+from exceptions import ParsingError, UnexpectedColumnCountError
+from model import ActionType, BrokerTransaction
+
+
+def action_from_str(label: str):
```
KapJI:
```suggestion
def action_from_str(label: str) -> ActionType:
```

KapJI (approved):

KapJI:
Looks great, thanks! 🎉

PR #12, opened 2021-04-12
Add trading212 support

vmartinv:
This adds the parser for Trading212 together with an example and instructions.

### DIFF
--- a/parsers.py
+++ b/parsers.py
@@ -8,6 +8,7 @@ from dates import date_to_index
 from exceptions import UnexpectedColumnCountError
 from model import BrokerTransaction, DateIndex
 from schwab import read_schwab_transactions
+from trading212 import read_trading212_transactions
 
 
 class InitialPricesEntry:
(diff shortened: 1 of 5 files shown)


parsers.py
```diff
@@ -27,9 +28,12 @@ def __str__(self) -> str:
...
 
 
-def read_broker_transactions(schwab_transactions_file: str) -> List[BrokerTransaction]:
+def read_broker_transactions(
+    schwab_transactions_file: str, trading212_transactions_folder: str
+) -> List[BrokerTransaction]:
     reader_transactions = [
         read_schwab_transactions(schwab_transactions_file),
+        read_trading212_transactions(trading212_transactions_folder),
     ]
```
KapJI, line 37: I think we shouldn't merge them. Otherwise you can end up with selling shares from one account on another account which will cause incorrect results. Let's chose which broken we want to use somewhere.

KapJI (changes requested):

parsers.py
vmartinv: The thing is that for tax calculations it should be merged. HMRC pool system needs to include all the shares you own regardless of where you obtained them. So if I want to calculate my gains I need to include the transactions from all the brokers. For example buying FOO in some broker and then selling FOO acquired two years ago in some other broker will count as Same day rule.

KapJI: Yeah, I think you're right. Ok, let's merge then.

KapJI (approved):

PR #13, opened 2021-04-12
Add report class

vmartinv:
Instead of just printing it encapsulates the results into an object. This will be useful for writing the tests.


KapJI (approved):
Awesome, thanks!

PR #14, opened 2021-04-13
move currency converter and initial prices into separate modules

vmartinv:
This reduces `calc.py` size. Next diff will remove some global variables.


KapJI:
Thanks!

PR #17, opened 2021-04-13
Add unit tests

vmartinv:
This adds some unit tests, and makes them run in the CI too. There are three tests, two of them correspond to examples given by the HMRC (links provided). More tests should be added to increase coverage.

Tried to add one integration test that runs the script with the example files but installing `pdflatex` in the CI doesn't seem very straightforward at all.

(I had to make one small fix, symbol should be null for transfer transactions...)

### DIFF
--- a/tests/test_calc.py
+++ b/tests/test_calc.py
@@ -226,7 +226,7 @@ def test_run_with_example_files():
     assert result.stderr == b"", "Run with example files generated errors"
     with open("tests/test_run_with_example_files_output.txt", "r") as file:
         expected = file.read()
-    assert result.stdout.decode('utf-8') == expected, (
+    assert result.stdout.decode("utf-8") == expected, (
         "Run with example files generated unexpected outputs, if you added new features update the test with:\n"
         + " python calc.py --report '' --schwab schwab_transactions.csv --trading212 trading212/"
         + " > tests/test_run_with_example_files_output.txt"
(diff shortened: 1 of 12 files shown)


KapJI:
Let's add integration test with disabled pdf generation. This will make sure that no asserts failed at least.

tests/test_calc.py
```diff
@@ -0,0 +1,203 @@
...
+    return calculator.calculate_capital_gain(acquisition_list, disposal_list)
+
+
+def buy_transaction(
+    date: datetime.date,
+    symbol: str,
+    quantity,
+    price,
+    fees,
+    amount,
```
KapJI: Can you add type hints where possible?

vmartinv:
added suggestions

tests/test_calc.py
```diff
@@ -0,0 +1,223 @@
...
+def test_run_with_example_files():
+    args = [
+        "--schwab",
+        "schwab_transactions.csv",
+        "--trading212",
+        "trading212/",
+        "--report",
+        "",
+    ]
+    subprocess.run(["python", "calc.py"] + args, check=True)
```
KapJI: Let's check the output maybe? At least for some key lines.

KapJI:
Thanks a lot for working on this! 🚀

KapJI (approved):

PR #18, opened 2021-04-14
improve integration test

vmartinv:
This test would have failed in one year, I forgot to pass the tax year. Also removes some redundancy.


KapJI:
Good spot! Thanks!

PR #56, opened 2021-05-09
Add support for Morgan Stanley

goofy57:


cgt_calc/parsers/mssb.py
```diff
@@ -0,0 +1,165 @@
...
+    "Tax Payment Method",
+]
+
+# These can be potentially wired through as a flag
+KNOWN_SYMBOL_DICT: Final[dict[str, str]] = {
+    "GSU Class C": "GOOG",
+}
+
+
+def __hacky_parse_decimal(decimal: str) -> Decimal:
```
KapJI: 2 underscores are usually used for some python special methods. If you want to mark it private, I think single underscore is fine.
```suggestion
def _hacky_parse_decimal(decimal: str) -> Decimal:
```

goofy57: Done.

KapJI (approved):
Thanks! 🎉

PR #75, opened 2021-06-19
Fixes and simplifications

ivendor:
-Fixed an edge case in calculation of bed and breakfast rule where  quantity of stock to be dispose is bigger than quantity of stocks that fells under B&B rule.
-Removed indirection of using integers for dates  as indices as in python dates object are already hashable


KapJI:
Can you please split this into 2 PRs?

PR #102, opened 2021-10-02
An additional assertion for B&B rule

ivendor:
just a minor thing on top of the recent fix you merged to make sure we aren't left with leftover quantity.

Also adding a repr to the entries for easy printing when debugging

### DIFF
--- a/cgt_calc/main.py
+++ b/cgt_calc/main.py
@@ -530,6 +530,9 @@ class CapitalGainsCalculator:
                     new_pool_cost=current_amount,
                 )
             )
+            disposal_quantity = Decimal(0)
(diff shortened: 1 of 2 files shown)


KapJI (approved):

KapJI:
There are some CI errors, please take a look.

cgt_calc/main.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python3
+cgt_calc/main.py#!/usr/bin/env python3
```
KapJI:
```suggestion
#!/usr/bin/env python3
```

PR #103, opened 2021-10-02
Simplifying abstraction by Using date as indices instead of numbers

ivendor:
As mentioned in the title we don't need to pass for an intermediate abstractions as dates are hashable and sortable already

### DIFF
--- a/tests/test_calc.py
+++ b/tests/test_calc.py
@@ -10,7 +10,6 @@ import sys
 import pytest
 
 from cgt_calc.currency_converter import CurrencyConverter
-from cgt_calc.dates import date_to_index
 from cgt_calc.initial_prices import InitialPrices
 from cgt_calc.main import CapitalGainsCalculator
 from cgt_calc.model import (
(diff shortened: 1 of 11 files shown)


KapJI (approved):
Thanks! 🙂

tests/test_calc.py
```diff
@@ -125,7 +124,7 @@ def transfer_transaction(
         1.00,  # Expected capital gain/loss
         None,
         {
-            date_to_index(datetime.date(day=1, month=5, year=2020)): {
+            (datetime.date(day=1, month=5, year=2020)): {
```
KapJI: No need in these brackets anymore.
```suggestion
            datetime.date(day=1, month=5, year=2020): {
```

ivendor:
Addressed all issues and re-tested manually too

tests/test_calc.py
KapJI: I think you missed this one. And there're more places like this below.

ivendor: Yep you're right, updated it now

PR #104, opened 2021-10-03
Schwab use awards centre report to calculate the stock price of the Stock Activities

ivendor:
For schwab we can now use the awards centre report to calculate the stock price of the Stock Activities.

Also fixed a double conversion to gbp bug in the stock activity calculation


tests/test_data/test_run_with_example_files_output.txt
```diff
@@ -21,15 +22,15 @@ Portfolio at the end of 2020/2021 tax year:
   GE: 1.00, £8.07
   NVDA: 1.00, £401.01
   FOO: 30.50, £543.97
-  GOOG: 8.60, £8908.20
+  GOOG: 8.60, £12516.92
```
ivendor, line 25: The dobule conversion problem is particularly visible here where

8.60 stocks *  2045.06 price / 1.4051 currency conversion should give 12516.92

PR #111, opened 2021-10-31
Better modeling Hmrc transactions

ivendor:
Use a more structured dataclass instead of loose tuples to model HMRC transactions and prevent errors for future changes


KapJI (approved):
Neat, thanks!

PR #112, opened 2021-10-31
Automated currency rate updater

ivendor:
This diff introduces a script to automatically update currency rates file from HRMC goverment data.

It still needs to be run manually but it could be integrated either in the CI automation or directly live in the software with the fallback on the offline file.


currency_updater.py
```diff
@@ -0,0 +1,139 @@
...
+        return cast(
+            TextIO,
+            urllib.request.urlopen(
+                urllib.request.Request(
+                    "http://www.hmrc.gov.uk/softwaredevelopers/rates/"
+                    f"exrates-monthly-{date_str}.XML",
+                    headers=http_headers,
+                )
+            ),
+        )
```
KapJI, line 76: It just retries ignoring exception? I think it's better to log the error and tell that request has failed.

KapJI (commented):
Please update typing to the new format.

KapJI:
Implemented by https://github.com/KapJI/capital-gains-calculator/pull/157.

PR #131, opened 2021-11-29
Add support for data from Sharesight service

danielkza:
Sharesight is a portfolio tracker that can aggregate data from multiple 
brokers. Add support for importing data from All Trade and Taxable Income
reports that the service provides.

Due to Sharesight not being a broker, reports don't include information 
about deposits and withdrawals; hence, checking for negative balances is 
not only not useful, but impossible. Add a command line option
`--no-balance-check` to handle this problem.

There are some additional small fixes include for issues I found while
developing. I can split those up if needed.

README and tests updated to match.


cgt_calc/parsers/sharesight.py
```diff
@@ -0,0 +1,284 @@
...
+    """Sharesight transaction.
+
+    Just a marker type for now
+    """
+
+
+class RowIterator(Iterator[List[str]]):
+    """Iterator for CSV rows that keeps track of line number."""
+
+    def __init__(self, rows: Iterable[list[str]]) -> None:  # noqa: D107
```
KapJI: IMO, trivial docstring like `Initialise RowIterator` is better than `noqa` comment.

cgt_calc/parsers/sharesight.py
```diff
@@ -0,0 +1,284 @@
...
+        self.rows = iter(rows)
+        self.line = 1
+
+    def __next__(self) -> list[str]:
+        """Produce next element and increment line number."""
+        elm = next(self.rows)
+        self.line += 1
+        return elm
+
+    def __iter__(self) -> RowIterator:  # noqa: D105
```
KapJI: Same here.

cgt_calc/parsers/sharesight.py
```diff
@@ -0,0 +1,284 @@
...
+def parse_dividend_payments(
+    rows: Iterator[list[str]],
+) -> Iterable[SharesightTransaction]:
+    """Parse dividend payments from Sharesight data.
+
+    This is the section started with a "Foreign Income" or "Local Income" header.
+    We parse those two sections very similarly, so we use one function.
+    """
+
+    columns = next(rows)  # pylint: disable=R1708
```
KapJI: Please add a small explanation for the comment. E.g. link to https://github.com/PyCQA/pylint/issues/3424

KapJI (commented):
Looks great, thanks! I've added few minor comments.

KapJI:
Also, please update example PDF to reflect template change.

cgt_calc/parsers/sharesight.py
danielkza: I relented and just accepted the lint suggestion :D

cgt_calc/parsers/sharesight.py
danielkza: fixed

cgt_calc/parsers/sharesight.py
danielkza: fixed

PR #155, opened 2022-01-26
relax trading212 validation, allow missing columns

vmartinv:
It seems that when a column is empty Trading212 might remove it entirely from the csv, causing the parser to fail. This relaxes the validation to fail only on unknown columns and can also handle the conversion fee which can sometimes appear.


KapJI:
Thanks!

Issue #57, opened 2021-05-09
Use HMRC API to fetch exchange rates


KapJI:
Embedded file with GBP/USD prices can be replaced with this API: https://developer.service.hmrc.gov.uk/api-documentation/docs/api/xml/Exchange%20rates%20from%20HMRC

KapJI:
Done by https://github.com/KapJI/capital-gains-calculator/pull/157

PR #157, opened 2022-01-28
Query exchange rates from HMRC API

vmartinv:
This updates the currency converter class by querying the HMRC API suggested in #57.

### Features
- It supports converting all available ​currencies (not only USD)
- Data is queried on demand
- It keeps a cache in memory
- Saves the data in a human-readable file for future runs
- Tested with the test suite and my personal data

Adding a small sleep can be considered to prevent querying too fast


cgt_calc/args_parser.py
```diff
@@ -58,11 +58,11 @@ def create_parser() -> argparse.ArgumentParser:
...
 
     parser.add_argument(
-        "--gbp-history",
+        "--exchange-rates-file",
         type=str,
-        default=None,
+        default=DEFAULT_EXCHANGE_RATES_FILE,
         nargs="?",
-        help="monthly GBP/USD prices from HMRC",
+        help="monthly exchange rates from HMRC",
```
KapJI:
```suggestion
        help="output file for monthly exchange rates from HMRC",
```

cgt_calc/currency_converter.py
```diff
@@ -1,37 +1,112 @@
...
+from xml.etree import ElementTree
+
+import requests
 
 from .dates import is_date
-from .exceptions import ExchangeRateMissingError
+from .exceptions import ExchangeRateMissingError, ParsingError
 from .model import BrokerTransaction
 
+EXCHANGE_RATES_HEADER = ["month", "currency", "rate"]
```
KapJI:
```suggestion
EXCHANGE_RATES_HEADER: Final = ["month", "currency", "rate"]
```

KapJI (commented):
Looks great, thank you!

PR #252, opened 2022-08-18
Show average unit price in calculations PDF.

thibwk:
This makes it easier to understand how the pool unit cost evolves, and makes it easier to understand the capital gain or losses.


KapJI:
Makes sense, thanks!

PR #277, opened 2022-10-10
Add a parser for Schwab Equity Award data

thibwk:
In particular this covers the Alphabet/Google equity awards, which data is in a different format than what the Schwab parser already available expects.
This parser also accounts for the Alphabet stock split given the inconsistent data that Schwab still provides (some data expressed in pre-split numbers, some in post-split).


KapJI:
Please fix merge conflict :)
Restore old version of `poetry.lock`, rebase and add dependency again.

KapJI:
Thanks!

PR #305, opened 2022-12-25
Increase decimal precision.

thibwk:
This is because a real-life example needed a fifth decimal, so increasing to 6 decimals to have some margin.

Also take the opportunity to use the common round_decimal() function instead of having a separate one in the Schwab Equity Award parser.


anthrotype (approved):

KapJI (approved):
Thanks!

PR #306, opened 2022-12-25
Extend sale parsing to support multiple subtransactions with different prices

thibwk:
As seen in real-life data.


anthrotype (approved):

KapJI (approved):
Thanks!

PR #399, opened 2023-06-25
Update Morgan Stanley parser

goofy57:
- Handle GOOG stock split correctly
- Treat "Staged" releases similarly to "Completed"


KapJI:
Thanks!

PR #411, opened 2023-09-02
Schwab award report dates should always takes precedence over activity dates

ivendor:
Schwab excel reporting for stock activities date reports erroneous dates sometimes, for example the next day respect to what is actually exported  in the PDF statements.

This can potentially interfere with UK same day acquisition selling rule or bed and breakfast.

For example you can receive an award on the 15th and sell on 16th of the same month and the xls for activities will report both on the 16th.


KapJI:
Thanks!

PR #443, opened 2023-12-03
Support "raw" csv for a simple data format.

nihn:
Suppport a "raw" csv which only contains minium set of columns for the calcualtion. Such CSV can be produced from any broker with a simple script, it's a generic version to support all different brokers.


KapJI:
Can you fix lint warnings please?

nihn:
I forgot how to use git :)

KapJI:
Thanks!

PR #465, opened 2024-01-18
Update Trading212 parser, keeping it backwards compatible

vmartinv:
There are minor changes in Trading212 format, is easy to update it.


KapJI (approved):
Thanks!

PR #467, opened 2024-01-20
Support updated JSON format in Schwab Equity Award parser

thibwk:
Schwab have now updated their website to:
-  allow the whole history of Equity Awards account to be downloaded (previously only the last 4 years were made available)
- via a simple Export button, in an updated JSON format (and CSV and XML).

This PR updates the Schwab Equity Award parser to work with this updated format (while still supporting the previous one).


cgt_calc/parsers/schwab_equity_award_json.py
```diff
@@ -216,9 +285,10 @@ def __init__(
...
 
                     quantity = (amount + fees) / price
+
         else:
             raise ParsingError(
-                file, f'Parsing for action {row["action"]} is not implemented!'
+                file, f"Parsing for action {row[names.action]} is not implemented!"
             )
 
         currency = "USD"
```
m01, line 295: Note: a few lines below (line 311+ in the new code/241+ in the old code - I can't add a comment there directly sadly) the `_normalize_split()` function seems to assume that the data is about Alphabet shares. I think some check on the stock symbol might be prudent to avoid accidentally scaling some other stock's values according to the Alphabet split, in cases where the user happens to also use Schwab as broker for employer-issued equity.

tests/test_data/schwab_equity_award_v2.json
```diff
@@ -0,0 +1,144 @@
...
+            "Description": "RS",
+            "FeesAndCommissions": null,
+            "DisbursementElection": null,
+            "Amount": null,
+            "TransactionDetails": [
+                {
+                    "Details": {
+                        "AwardDate": "01/01/2019",
+                        "AwardId": "C987654",
+                        "VestDate": "09/25/2023",
```
tiziano88, line 19: what date format on earth is this? ah yes, US...

thibwk: Yes ISO 8601 date formats would be clearer and safer but we'll have to work with what Schwab gives us!

cgt_calc/parsers/schwab_equity_award_json.py
thibwk: Yes that's right, that's for Alphabet's stock split. I am not sure if Schwab gives the same data format as parsed by this code for other accounts than the Google Equity Awards ones which can only have GOOG and GOOGL. 
But I agree it'd be better to also check on the stock ticker name to be safe.

Not directly relevant to this PR though.

thibwk:
> To me it seems we have two parsers for the same thing, in schwab.py there's `_read_schwab_awards` which parses the csv format of this new export (instead of json). If they are equivalent I'd rather have the csv format.

Fair question, but the new CSV file format that Google employees get on their Equity Awards Account appear to have a different format, not supported by this parser.
e.g. the headers I see is:

```
Date	Action	Symbol	Description	Quantity	FeesAndCommissions	DisbursementElection	Amount	AwardDate	AwardId	VestDate	VestFairMarketValue	Type	Shares	SalePrice	SubscriptionDate	SubscriptionFairMarketValue	PurchaseDate	PurchasePrice	PurchaseFairMarketValue	DispositionType	GrantId	GrossProceeds
```

That doesn't seem to match https://github.com/KapJI/capital-gains-calculator/blob/2beffd677e941c20eb56e71c7f84f5ff0cd2f1c1/cgt_calc/parsers/schwab.py#L183.

In other words, Schwab seems to export data in different formats for different types of accounts?

m01:
> In other words, Schwab seems to export data in different formats for different types of accounts?

I thought the CSV parser was for "normal" individual brokerage accounts, and the equity awards JSON parser was for equity awards accounts.

vmartinv:
> That doesn't seem to match
> 
> https://github.com/KapJI/capital-gains-calculator/blob/2beffd677e941c20eb56e71c7f84f5ff0cd2f1c1/cgt_calc/parsers/schwab.py#L183
> 

No, that’s just the normal transaction parser. There are two parsers in schwab.py, the award parser is in the  function ‘_read_schwab_awards’

cgt_calc/parsers/schwab_equity_award_json.py
m01: The equity awards accounts and this JSON format are not unique to GOOG/GOOGL stocks, but I think this JSON format may be unique to equity awards accounts.

KapJI: It'd be great actually to have some JSON/CSV file in resources with information about splits to have general support for them instead of hardcoding a specific one directly in the code.

KapJI:
Thanks for finding how to export more data!

PR #469, opened 2024-01-23
Fix bug with balance not getting updated for action REINVEST_SHARES

amaramrahul:


vmartinv:
can you add a test in test_calc.py with this action?

amaramrahul:
I have just realized that my diff can be simplified by grouping the actions of ActionType.BUY and ActionType.REINVEST_SHARES. I guess this obliviates the need for additional test cases.

KapJI:
Thanks!

PR #471, opened 2024-01-23
Dockerfile

deed02392:
add basic dockerfile that drops into a bash shell with cgt-calc on path from a pipx venv


vmartinv:
What do you use this for? Could you also add a section in the README with some commands?

deed02392:
@vmartinv this Docker container means you don't have to install LaTeX on your host (which can take a while). I made this Dockerfile with all the necessary dependencies, so that if you want, you could officially link this repo with say hub.docker.com and then people could simply `docker pull capital-gains-calculator` and not have to worry about separate build instructions etc..

I updated the README.md with instructions on how it works. The gist is you get the tool running in the container and then just share your tax files with it, run the tool and the output PDF gets dropped in the same directory on your host. You could even generate the report with a single command, rather than get an interactive shell in the container first.

Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+    wget
+
+
+RUN luaotfload-tool --update
+RUN apk --no-cache add py3-pandas
+RUN apk --no-cache add pipx
+RUN pipx install cgt-calc
+RUN pipx ensurepath
+# workaround https://github.com/KapJI/capital-gains-calculator/issues/470
+RUN pipx inject cgt-calc pandas
```
KapJI: This should be no longer needed.

deed02392: rebuilt, tested, confirmed

KapJI:
Please format README with prettier to pass CI.

KapJI:
Thanks!

PR #485, opened 2024-02-06
Better comments for raw format

deed02392:
Improve comments


KapJI:
Can you run formatters locally? It looks you have trailing whitespaces somewhere.

PR #524, opened 2024-06-02
Fix spin-offs calculations

nihn:
This diff fixes Spin-Off cost calculations (both for source and dest shares). Details of the calculations can be found in the comment of the main method. As we do not have a good DB for spin-offs the "source" share has to be provided by user during the run, the adjustments to the prices are then explained in generated `.pdf` file.

Additionally:
* Adding few missing transaction types to Schwab parser.
* Better error handling here and there.
* Move transactions (disposals, acquisitions, portfolios) to the instance variables to avoid passing them around all the time and making the class easier to understand (IMHO).

I'll add tests after initial review round.

### DIFF
--- a/cgt_calc/main.py
+++ b/cgt_calc/main.py
@@ -120,7 +120,7 @@ class CapitalGainsCalculator:
             if price is None:
                 price = self.initial_prices.get(transaction.date, symbol)
             amount = round_decimal(quantity * price, 2)
-        if transaction.action is ActionType.SPIN_OFF:
+        elif transaction.action is ActionType.SPIN_OFF:
             price, amount, spin_off = self.handle_spin_off(transaction)
         elif transaction.action is ActionType.STOCK_SPLIT:
             price = Decimal(0)
@@ -155,10 +155,11 @@ class CapitalGainsCalculator:
         Doc basing on SOLV spin off out of MMM.
 
         # 1. Determine the Cost Basis (Acquisition Cost) of the SOLV Shares
-        In the UK, the cost basis (or acquisition cost) of the new SOLV shares received
-        from the spin-off needs to be determined. This is usually done by apportioning
-        part of the original cost basis of the MMM shares to the new SOLV shares based
-        on their market values at the time of the spin-off.
+        In the UK, the cost basis (or acquisition cost) of the new SOLV shares
+        received from the spin-off needs to be determined. This is usually done
+        by apportioning part of the original cost basis of the MMM shares to
+        the new SOLV shares based on their market values at the time of the
+        spin-off.
 
         ## Step-by-Step Allocation
         * Find the Market Values:
@@ -168,25 +169,25 @@ class CapitalGainsCalculator:
 
         * Calculate the Apportionment:
 
-        Divide the market value of the MMM shares by the total market value of both MMM
-        and SOLV shares to find the percentage allocated to MMM.
+        Divide the market value of the MMM shares by the total market value of
+        both MMM and SOLV shares to find the percentage allocated to MMM.
         Do the same for SOLV shares to find the percentage allocated to SOLV.
 
         * Allocate the Original Cost Basis:
 
-        Multiply the original cost basis of your MMM shares by the respective percentages
-        to allocate the cost basis between the MMM and SOLV shares.
+        Multiply the original cost basis of your MMM shares by the respective
+        percentages to allocate the cost basis between the MMM and SOLV shares.
 
         ## Example Allocation
         * Original Investment:
 
-        Assume you bought 100 shares of MMM at £100 per share, so your total cost basis
-        is £10,000.
+        Assume you bought 100 shares of MMM at £100 per share, so your total
+        cost basis is £10,000.
 
         * Market Values Post Spin-off:
 
-        Assume the market value of MMM shares is £90 per share and SOLV shares is £10 per
-        share immediately after the spin-off.
+        Assume the market value of MMM shares is £90 per share and SOLV shares
+        is £10 per share immediately after the spin-off.
         The total market value is £90 + £10 = £100.
 
         * Allocation Percentages:
@@ -201,8 +202,8 @@ class CapitalGainsCalculator:
 
         # 2. Determine the Holding Period
         The holding period for the SOLV shares typically includes the holding
-        period of the original MMM shares. This means the date you acquired the MMM shares
-        will be used as the acquisition date for the SOLV shares.
+        period of the original MMM shares. This means the date you acquired the
+        MMM shares will be used as the acquisition date for the SOLV shares.
         """
         symbol = transaction.symbol
         quantity = transaction.quantity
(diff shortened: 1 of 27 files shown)


cgt_calc/main.py
```diff
@@ -276,107 +391,92 @@ def convert_to_hmrc_transactions(
...
-        assert acquisition_quantity >= 0
+        assert acquisition.quantity >= 0
         # Stock split can have 0 amount
-        assert acquisition_amount >= 0
-        bed_and_breakfast_quantity = Decimal(0)
-        bed_and_breakfast_amount = Decimal(0)
+        assert acquisition.amount >= 0
+
+        bnb_acquisition = HmrcTransactionData()
         bed_and_breakfast_fees = Decimal(0)
```
KapJI, line 398: This can be taken from `bnb_acquisition` as well.

cgt_calc/main.py
```diff
@@ -276,107 +391,92 @@ def convert_to_hmrc_transactions(
...
         chargeable_gain = Decimal(0)
         calculation_entries = []
         # Same day rule is first
-        if has_key(acquisition_list, date_index, symbol):
-            same_day_quantity, same_day_amount, _same_day_fees = astuple(
-                acquisition_list[date_index][symbol]
-            )
-            available_quantity = min(disposal_quantity, same_day_quantity)
+        if has_key(self.acquisition_list, date_index, symbol):
+            sd_acquisition = self.acquisition_list[date_index][symbol]
```
KapJI: I'm ok with `bnb` for `bed_and_breakfast` but this one hurts readability IMO.

```suggestion
            same_day_acquisition = self.acquisition_list[date_index][symbol]
```

cgt_calc/main.py
```diff
@@ -409,37 +509,30 @@ def process_disposal(
...
+                    acquisition = self.acquisition_list[search_index][symbol]
+
+                    bnb_acquisition = (
+                        self.bnb_list[search_index][symbol]
+                        if has_key(self.bnb_list, search_index, symbol)
+                        else HmrcTransactionData()
+                    )
+                    assert bnb_acquisition.quantity <= acquisition.quantity
+
+                    sd_disposal = (
```
KapJI:
```suggestion
                    same_day_disposal = (
```

cgt_calc/main.py
```diff
@@ -276,107 +391,92 @@ def convert_to_hmrc_transactions(
...
         date_index: datetime.date,
     ) -> tuple[Decimal, list[CalculationEntry]]:
         """Process single disposal."""
-        disposal_quantity, proceeds_amount, disposal_fees = astuple(
-            disposal_list[date_index][symbol]
-        )
+        disposal = self.disposal_list[date_index][symbol]
+        disposal_quantity = disposal.quantity
+        proceeds_amount = disposal.amount
+        disposal_fees = disposal.fees
```
KapJI: Only used once and not modified, can be used as `disposal.fees`

cgt_calc/main.py
```diff
@@ -409,37 +509,30 @@ def process_disposal(
                         fees=fees,
                         new_quantity=current_quantity,
                         new_pool_cost=current_amount,
+                        spin_off=None,
```
KapJI: It's default to `None`.
```suggestion
```

cgt_calc/main.py
```diff
@@ -127,40 +136,149 @@ def add_acquisition(
...
+        price = self.price_fetcher.get_closing_price(symbol, transaction.date)
+        amount = quantity * price
+
+        while True:
+            # This would ideally be fetched from some stock DB but yfinance does not
+            # provide any info on SpinOffs
+            ticker = input(
+                "For a spin off, please enter the original ticker from which the new "
+                "stock was spinned off: "
+            )
```
KapJI: How user will understand what split this is about?

cgt_calc/main.py
```diff
@@ -127,40 +136,149 @@ def add_acquisition(
...
+            if ticker in self.portfolio:
```
KapJI: Maybe show the list of options as well?

cgt_calc/main.py
```diff
@@ -127,40 +136,149 @@ def add_acquisition(
...
+        assert symbol is not None
+        assert quantity is not None
+
+        price = self.price_fetcher.get_closing_price(symbol, transaction.date)
+        amount = quantity * price
+
+        while True:
+            # This would ideally be fetched from some stock DB but yfinance does not
+            # provide any info on SpinOffs
+            ticker = input(
```
KapJI: Can you add option to provide required info from text file? To make it possible to rerun script without specifying any input. Ideally, prices from yfinance can be stored there as well. E.g. here after fetching you can store data in some cache file and check its presence.

cgt_calc/main.py
```diff
@@ -127,40 +136,149 @@ def add_acquisition(
...
+
+        total_amount = sum(
+            t.amount
+            for transactions in self.acquisition_list.values()
+            if (t := transactions.get(ticker))
+        ) - sum(
+            t.amount
+            for transactions in self.disposal_list.values()
+            if (t := transactions.get(ticker))
+        )
```
KapJI: Why not use current value from portfolio?

cgt_calc/main.py
```diff
@@ -127,40 +136,149 @@ def add_acquisition(
...
+        share_of_original_cost = total_amount / (amount + total_amount)
```
KapJI: Why total amount increased here (`amount + total_amount`)?

KapJI (commented):
Thanks for implementing this! It needs some test to make sure the calculations are correct.

cgt_calc/main.py
nihn: Good point :D

cgt_calc/main.py
nihn: good point, this is left over from a previous, invalid version of this diff

cgt_calc/main.py
nihn: That may be a long list, I guess we can show a trimmed one.

cgt_calc/main.py
nihn:
```
Assume the market value of MMM shares is £90 per share and SOLV shares
is £10 per share immediately after the spin-off.
The total market value is £90 + £10 = £100.
```
we need to consider cost basis of "spin-off pool" as value of original stock and new stock.

cgt_calc/main.py
nihn: in the original code it was recalculated as 
```
                bed_and_breakfast_fees = (
                    acquisition_fees * bed_and_breakfast_quantity / acquisition_quantity
                )
```
I do not know enough on BnB to make a judgment if it will be the same, if you confirm I am fine with that :)

KapJI: Right, let's keep it then.

KapJI:
Can you rebase on the latest revision?

nihn:
I don't think it will change anything for that B&B case as I am not touching stock splits but B&B should be handled correctly in case of spin-offs (AFAICT, as it just modifies the pool price, I am not tax advisor though :)).

nihn:
Any idea why the signals are failing? I pulled the branch and it works on my machine:
image

vmartinv:
> Any idea why the signals are failing? I pulled the branch and it works on my machine

You're not including the spin_off_handler file in the commit.

nihn:
> You're not including the spin_off_handler file in the commit.

How did it work before is beyond my understanding though, thanks! :D

PR #531, opened 2024-06-07
Update and fix Trading212 parser

vmartinv:
This introduces a few improvements to the Trading212 parser:
- Adds support for a few new columns (keeping backward compatibility with the old format)
- Fixes a bug when calculating the fees (I was able to spot this when I compared the calculated balance vs the actual balance in my account). Trading212 rounds off the price strangely so I just added an assert that checks that the discrepancy is less than 1 cent per share.
- Added one test with the new format


cgt_calc/parsers/trading212.py
```diff
@@ -98,32 +105,57 @@ def __init__(self, header: list[str], row_raw: list[str], filename: str):
...
             currency = "GBP"
+        if (
+            amount is not None
+            and (action == ActionType.BUY or self.raw_action == "Withdrawal")
+            and amount > 0
+        ):
+            amount *= -1
         price = (
-            abs(amount / quantity)
+            abs((amount + fees) / quantity)
```
KapJI: If amount is negative but fees are always positive?
```suggestion
            (abs(amount) + fees) / quantity)
```

cgt_calc/parsers/trading212.py
```diff
@@ -98,32 +105,57 @@ def __init__(self, header: list[str], row_raw: list[str], filename: str):
...
-                amount *= -1
-            amount -= fees
+        reported_price = self.price_foreign
+        if (
+            reported_price is not None
+            and price is not None
+            and (self.currency_foreign == "GBP" or self.exchange_rate is not None)
+        ):
+            reported_price /= self.exchange_rate or Decimal("1")
+            assert abs(reported_price - price) < Decimal("0.01"), (
```
KapJI: That could be a reason why you had mismatch here.

cgt_calc/parsers/trading212.py
vmartinv: No, that won't work. The formula is to remove the fees from the total paid/received.
- If purchasing 1 share costs £10 with £1 fee, price / share is abs(-10+1)/1 = £9
- If selling 1 share gives me £9 with £1 fee, price / share is abs(9+1)/1 = £10

cgt_calc/parsers/trading212.py
vmartinv: The reason there's a mismatch is because trading212 rounds the price / share.
Even when checking the transaction in the UI I get this discrepancy:
Sell 24 shares at $164.01 with exchange rate 1.25229 and £4.71 fees. Trading212 total says £3138.50. However,
24*164.01/1.25229-4.71=£3138.52

I believe there might be some implicit rounding both in the csv and in the UI. Some of them I was able to fetch more decimals from the UI and updating the csv. Others are unfixable, but since is just a few cents is not that important. I set it to a warning now when the discrepancy is more than 0.015 in the share price.
Ultimately, what matters is the £££ received/sent, which is what I'm using here.

KapJI:
Looks good to me, thanks!

PR #532, opened 2024-06-09
Add basic support for cash merger transactions (Schwab only)

vmartinv:
This is a rudimentary implementation to handle cash merger transactions (i.e. when a company is acquired by another one). In these events the shares of the original company cease to exist and the investor can receive cash and shares from the new company in return.

In my case I only received cash, this case in simpler in that it is just like a normal sell event, which is what I implemented here.
When you also receive shares, the calculation is more complex and there's an election on how you want it to be taxed. I can't implement this as I don't have any past experience or data for it but more details are given in the HRMC guidance:
https://www.gov.uk/guidance/capital-gains-tax-share-reorganisation-takeover-or-merger#shares-and-cash

## Implementation

- In the Schwab csv export, this transaction uses two rows so I added an additional pass in the parser two merge them.
- Includes asserts checking for the case implemented
- Contains a further a warning mentioning the lack of full support when this transaction is detected.
- Adds a test case

### DIFF
--- a/cgt_calc/main.py
+++ b/cgt_calc/main.py
@@ -205,7 +205,7 @@ class CapitalGainsCalculator:
             ]:
                 new_balance += get_amount_or_fail(transaction)
                 self.add_acquisition(portfolio, acquisition_list, transaction)
-            elif transaction.action is ActionType.SELL:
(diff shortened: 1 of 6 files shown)


KapJI:
Thanks for implementing this!

Issue #510, opened 2024-05-04
Support for "gift to spouse" action type


jezevec:
Is there a consideration to support gifts to spouse as a separate action?

I guess standard gifts are equivalent to selling shares. But giving to a spouse, despite not having tax implications, reduces Section 104. I don't see how to input that transaction at least as a raw input.

Skipping that transaction would lead to incorrect results causing this application cannot be used in such scenarios.

vmartinv:
I think it would be an useful feature to support. Although I believe you need to handle both sending and receiving the shares, it's likely both people will be using the tool.
1. Sending should be easy to implement as it just reduces the 104, although more careful consideration might be needed when the other rules come into scope. It might be helpful to report the value of the gift too for the reason below.
2. Receiving is harder to implement, because the giftee needs to somehow input the value of the shares when they were first  owned by the gifter.

I don't have transaction of these sorts so any help for implementing this would be appreciated.

Related: https://www.gov.uk/capital-gains-tax/gifts

jezevec:
I was looking into that and came to the same conclusions. Transfering to spouse should not be a subject to matching rules, only reduces the Section 104 (based on https://community.hmrc.gov.uk/customerforums/cgt/cd55f761-e320-ef11-a81c-0022481b5e57). I have created a patch for that. I have such a transaction but only use the raw format so haven't added it for other broakers.

Regarding receiving shares I am not fully clear whether it is subject to the same day and B&B rules. Some time ago I saw a post on the forum which implied it is, but when I checked again now that post is gone, so maybe it was wrong. Otherwise getting the quantity and value will be easy as it will be the same as the quantity and value that was removed from partner's Section 104.

anthrotype:
> Transfering to spouse should not be a subject to matching rules, only reduces the Section 104 (based on https://community.hmrc.gov.uk/customerforums/cgt/cd55f761-e320-ef11-a81c-0022481b5e57)

Are you 100% sure about that?

In another popular tool to compute CGT they say the following about transfers to spouse:

https://www.cgtcalculator.com/example7.htm

> When your shares are transferred to a spouse the shares are matched up in the same way as if the shares were sold. Your spouse acquires your shares for the same amount as the original acquisition cost. The result is zero gain in your capital gains calculations. For more see this [post](http://uk.advfn.com/cmn/fbb/thread.php3?id=24977116&from=455).

[...]

PR #539, opened 2024-06-30
Add parsing support for gift transactions (Schwab Equity Awards JSON).

m01:
This adds a new action type, as TRANSFER seems to deal with money transfers rather than share transfers, and SALEs also involve money whereas gifted shares do not.

The calculation code will raise a NotImplementedError if it encounters one of these transactions.

This may help #510. This contribution has been developed in my spare time.

### DIFF
--- a/cgt_calc/main.py
+++ b/cgt_calc/main.py
@@ -305,6 +305,10 @@ class CapitalGainsCalculator:
                 self.add_disposal(transaction)
                 if self.date_in_tax_year(transaction.date):
                     total_sells += self.converter.to_gbp_for(amount, transaction)
+            elif transaction.action is ActionType.GIFT:
+                raise NotImplementedError(
+                    f"{transaction.action} is not currently supported ({transaction})"
+                )
             elif transaction.action is ActionType.FEE:
                 amount = get_amount_or_fail(transaction)
                 new_balance += amount
--- a/cgt_calc/parsers/schwab_equity_award_json.py
+++ b/cgt_calc/parsers/schwab_equity_award_json.py
@@ -148,6 +148,9 @@ def action_from_str(label: str) -> ActionType:
     if label == "Wire Funds Received":
         return ActionType.WIRE_FUNDS_RECEIVED
 
+    if label == "Gift":
+        return ActionType.GIFT
+
     raise ParsingError("schwab transactions", f"Unknown action: {label}")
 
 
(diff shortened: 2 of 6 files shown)


cgt_calc/parsers/schwab_equity_award_json.py
```diff
@@ -199,6 +202,7 @@ def __init__(self, row: JsonRowType, file: str, field_names: FieldNames) -> None
         quantity = _decimal_from_number_or_str(row, names.quantity)
         amount = _decimal_from_number_or_str(row, names.amount)
         fees = _decimal_from_number_or_str(row, names.fees)
+        currency = "USD"
```
KapJI: Why have you moved it?

cgt_calc/main.py
```diff
@@ -305,6 +305,10 @@ def convert_to_hmrc_transactions(
                 self.add_disposal(transaction)
                 if self.date_in_tax_year(transaction.date):
                     total_sells += self.converter.to_gbp_for(amount, transaction)
+            elif transaction.action is ActionType.GIFT:
+                raise NotImplementedError(
```
KapJI, line 309: What's the point of adding it if it errors here?

m01: I thought `NotImplementedError` was a more appropriate error message than the `InvalidTransactionError` that IIRC it otherwise produces (as the transaction itself is valid), so it would be easier for users to understand why the tool isn't producing calculations.

cgt_calc/parsers/schwab_equity_award_json.py
m01: I wanted to set `currency` to `None`, but then realised that it's a (non-optional) `str` in `BrokerTransaction` (https://github.com/KapJI/capital-gains-calculator/blob/main/cgt_calc/model.py#L83), so I reverted that change and forgot to undo this change to where the default is specified. I'm happy to undo this move.

KapJI: Yeah, let's undo.

vmartinv:
If adding proper handling for gifts is out of the table, I think a better fix would be to raise an unknown transaction error in the parser if clarity is your concern. I see no need to add parsing for an unsupported transaction.
If a new contributor decides to implement gifts for a different parser they will have to worry about them working with Schwab which adds more work.

cgt_calc/parsers/schwab_equity_award_json.py
m01: done, or well, undone.

m01:
> If adding proper handling for gifts is out of the table,

It's not off the table - I plan to look into it but I don't really want to promise that I'll end up implementing it either at this stage.

>  If a new contributor decides to implement gifts for a different parser they will have to worry about them working with Schwab which adds more work.

I thought this would help progress support for gifts by one step, but also I don't mind if you want to hold off merging this until there's code using it for more than just error handling.

m01:
So https://github.com/KapJI/capital-gains-calculator/pull/549 adds support for a `TRANSFER_TO_SPOUSE` action type.

This PR and #549 could be linked up (and based on local testing it seems to work). The question is, how should gift transactions be identified as being to a spouse (as other types of gift may be taxed differently)? The data export doesn't contain enough data to determine that.

Possible approaches I've thought about:

* Assume all gift transactions are transfers to a spouse, possibly only if some flag is supplied.
* Depend on some extra data somewhere (maybe in a simple CSV file) that identifies transfer to spouse transactions.

I'm not sure what the best approach is for this project.

KapJI:
Closing, #924 covers this. Gift rows parse now, and the calculator asks who received them rather than failing on an unknown action. Gifts to a spouse are handled; gifts to anyone else are tracked in #932.

PR #541, opened 2024-07-02
Test for splits

sygi:
(Failing, do not merge) test showing wrong handling for stock splits.

### DIFF
--- a/tests/test_data/calc_test_data.py
+++ b/tests/test_data/calc_test_data.py
@@ -89,6 +89,26 @@ def transaction(
     )
 
 
+def split_transaction(
+    date: datetime.date,
+    symbol: str,
+    quantity: float,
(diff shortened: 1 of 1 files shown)


sygi:
I'm afraid I might need some support with this. In my mind, splits are basically an accounting trick: what was considered 2 stock yesterday is 10 units today, but there is no other effect. I tried to implement this above by changing the past acquisitions/disposals to operate in post-split space, but this leads to issues about pre-split reporting (if you sold one of the two units before split, it's 1 unit, not 5). This has been used in `mssb.py` so far.

One alternative I see is to enhance the `HmrcTransactionData` to keep an information about being ignored for matching purposes but then we'd need to remember about the conversion whenever we cross the line.

WDYT?

Issue #530, opened 2024-06-06
Error "texmf-dist-science (no such package)" when building docker image


jfresen:
In a fresh checkout, I'm following the instructions to run the docker image. But when executing `docker buildx build --platform linux/amd64 --tag capital-gains-calculator .`, I get an error `texmf-dist-science (no such package)`.

I'm not familiar with docker, but from what I can see, the package [doesn't exist on Alpine](https://pkgs.alpinelinux.org/packages?name=texmf-dist-science&branch=edge&repo=&arch=&maintainer=). Should this be fetched from somewhere else?

See [output.txt](https://github.com/user-attachments/files/15687237/output.txt) for the output of the command.

azatZul:
I have the same issue

PR #542, opened 2024-07-04
Pin Alpine version to 3.19

jfresen:
texmf-dist-science is not available on Alpine "edge" branch, nor on v3.20. The latest version where it is available is 3.19, so use that.

fixes KapJI/capital-gains-calculator#530


KapJI:
Thanks!

PR #549, opened 2024-07-28
Support "transfer to spouse" transaction

### LINKED ISSUE #510: Support for "gift to spouse" action type

jezevec:
It add support for transfers to spouse that have no GCT implications (if legally married and living together). It just reduces the Section 104.

This should fix KapJI#510


vmartinv:
It's important that we show the allowable cost for these shares, so that the spouse can use it for their tax calculations. I have two suggestions:
1. Add a info log mentioning the cost for each of the transfer transactions.
2. The PDF report is missing the symbol, see screenshot below. 
Screenshot 2024-12-24 at 10 52 30


We could even write an additional file with these costs so the spouse can load them , but we can do that in later PRs.

cgt_calc/main.py
```diff
@@ -768,6 +795,25 @@ def calculate_capital_gain(
...
+                    transfer_entry = CalculationEntry(
+                        rule_type=RuleType.TRANSFER_TO_SPOUSE,
+                        quantity=transaction.quantity,
+                        amount=-amount,
+                        new_quantity=self.portfolio[symbol].quantity,
+                        new_pool_cost=self.portfolio[symbol].amount,
+                        fees=transaction.fees,
+                        allowable_cost=amount,
+                    )
+                    calculation_log[date_index][f"transfer-to-spouse${symbol}"] = [
```
superbobry, line 814: You probably only want to do this if `date_index >= tax_year_start_index`, right?

KapJI:
Closing now that #924 is in. It uses the same action name and RAW format as this PR, plus the matching rules discussed on #510. Thanks again for the patch, it's what got this moving.

PR #563, opened 2024-09-28
Build cgt-calc inside Docker container instead of installing from pipx

justpresident:
This allows development without installing any tools locally.

Test plan:
```
docker build --tag capital-gains-calculator .
docker run --rm -it -v "$PWD":/data capital-gains-calculator:latest
e578f6a392e2:/data# cgt-calc --help
```


Dockerfile
```diff
@@ -17,15 +17,26 @@ RUN apk --no-cache add \
...
 
-WORKDIR /data
+WORKDIR /build
+
+RUN curl -sSL https://install.python-poetry.org | python3 -
+RUN ln -s /root/.local/share/pypoetry/venv/bin/poetry /bin/
+COPY cgt_calc /build/cgt_calc
+COPY .git /build/.git
+COPY .github /build/.github
+COPY pyproject.toml README.md .editorconfig .gitignore .kodiak.toml .pre-commit-config.yaml /build/
```
KapJI: Why don't copy everything or clone from github? Or mount this folder inside docker, so changes inside the container are available in the outside repo.

Dockerfile
```diff
@@ -17,15 +17,26 @@ RUN apk --no-cache add \
...
 RUN apk --no-cache add py3-pandas
-RUN apk --no-cache add pipx
-RUN pipx install cgt-calc
-RUN pipx ensurepath
 
-WORKDIR /data
+WORKDIR /build
+
+RUN curl -sSL https://install.python-poetry.org | python3 -
+RUN ln -s /root/.local/share/pypoetry/venv/bin/poetry /bin/
```
KapJI, line 31: I'm going to replace poetry and pipx with [uv](https://github.com/astral-sh/uv) soon.

KapJI (changes requested):

Dockerfile
justpresident: What difference does it make? I can't stop wondering to the wide variety of available and popular automatic packing tools in python that keep pulling some code from somewhere, compile and run it on my laptop!

Don't forget to update this Dockerfile thereafter then

Dockerfile
justpresident: > Why don't copy everything?

Yes, let's do that.

> or clone from github?

because You want to build from your changes.

> Or mount this folder inside docker, so changes inside the container are available in the outside

This is kind of the point - that I don't want all the changes to be outside of the container =) If it was inside a corporate code - makes sense, but for a public repository I would vote for more isolation.

justpresident:
I suppose it can be merged now?

Dockerfile
```diff
@@ -17,15 +17,23 @@ RUN apk --no-cache add \
...
 RUN luaotfload-tool --update
 RUN apk --no-cache add py3-pandas
-RUN apk --no-cache add pipx
-RUN pipx install cgt-calc
-RUN pipx ensurepath
 
-WORKDIR /data
+WORKDIR /build
+
+RUN curl -sSL https://install.python-poetry.org | python3 -
```
m01, line 30: What do you think about installing `poetry` using `apk` instead?

vmartinv: I prefer apk, it's more clear and easier to maintain.

PR #566, opened 2024-09-29
Fix transaction date in schwab_equity_award_json.

thibwk:
The previous logic was actually incorrect, as confirmed by comparing the output of cgt-calc with historical prices at https://www.nasdaq.com/market-activity/stocks/goog/historical.


thibwk:
Replacing this PR from `main` by [pull/569](https://github.com/KapJI/capital-gains-calculator/pull/569) which is the same change, but from a branch.

PR #568, opened 2024-10-30
Schwab Equity Award: support dividends and their taxes.

thibwk:
Google recently started paying dividends, which means Schwab uses new transaction types. Adding support for those.


vmartinv:
question: why are you using json instead of the normal export that schwab provides? (you can read using `--schwab-award`)

thibwk:
> question: why are you using json instead of the normal export that schwab provides? (you can read using `--schwab-award`)

The short answer is that the format of the CSV export for "Equity Award" accounts that Google/Alphabet employees get is different than what the capital-gains-calculator CSV Schwab parser expects. For instance, it doesn't have a `Price` column, and instead it has `SalePrice` and `VestFairMarketValue`.

Up until early 2024, they had horrible sub-heading within the CSV exports (i.e. headers mixed with data), making it very hard to work with. So we had to use JSON data then, and it was simpler to continue doing so (including for backward compatibility, for people who already downloaded their data as JSON in the past).

If we started from scratch today, we could probably use their new Equity Awards CSV format, but it isn't completely obvious: the data is still sparse (each line only has a few columns populated), and broken down by Equity award (e.g. each vest date, there are multiple lines, one per award).

PierfrancescoSoffritti:
Any plan to get this merged?

tests/test_schwab_equity_award_json.py
```diff
@@ -93,41 +93,73 @@ def test_schwab_transaction_v2() -> None:
...
+
+    i += 1
+    assert transactions[i].date == datetime.date(2023, 4, 25)
+    assert transactions[i].action == ActionType.STOCK_ACTIVITY
+    assert transactions[i].quantity == Decimal("13.6")
+    assert transactions[i].price == Decimal("106.78")
+    assert transactions[i].fees == Decimal("0")
+
+    i += 1
+    assert transactions[i].date == datetime.date(2023, 8, 29)
```
PierfrancescoSoffritti: Looks like tests are failing because of a typo in the date. It should be `datetime.date(2023, 8, 31)`

vmartinv:
> Why the preference of CSV over JSON?

Because:
1. The JSON format is meant for Schwab’s internal use, not for users, so it can change at any time without notice. It is truly more fragile.
2. Accessing the JSON takes extra effort specially if you lack developer knowledge, while downloading the CSV is quick, stable, and supported directly on their platform. It just seems like an unnecessary workaround.
3. The CSV is the official format Schwab provides for transactions, meaning it’s more likely to stay consistent over time.  
4. Even if the CSV isn’t an standard CSV, it’s still easy to parse or fix. I disagree about the fragility.

vmartinv:
> > Anyways, with that said there are some errors with this PR now, can you help resolving?
> 
> Sorry, I have been very busy, but I'll take a look as soon as I find a bit of time.
> 
> As for your last message about CSV vs. JSON: actually no, Schwab now offers the JSON file download directly from their website, making it as official and supported as the CSV... So I don't think there is anything blocking with JSON, is there?
> 
> (note that now their CSV format no longer has the crazy nested headers it used to have, I think I would try using CSV rather than JSON if I was starting today, mostly to make it easier for people to look at / edit the CSV / merge multiple files manually if they had to. But I haven't seen any strong-enough reason to spend time rewriting the whole parser yet...)

You provide two additional reasons to use CSV here. I personally never had to use JSON, I started with just CSV and it worked well, as you said is easy to merge and read, validate it contains all dates, etc. Is the reason all other parsers use CSV. If you still want to fix this parser I can accept it anyways.

PR #569, opened 2024-10-30
Fix sale dates in schwab_equity_award_json.

thibwk:
The previous logic was actually incorrect, as confirmed by comparing the output of cgt-calc with historical prices at https://www.nasdaq.com/market-activity/stocks/goog/historical.

This is because Schwab provides the trades date (which are the ones that matter for HMRC) and not the settlement dates as previously thought.


KapJI:
Thanks for figuring that out!

PR #571, opened 2024-11-17
Fix: add _approx_equal_price_rounding to allow price in accepable range

hermanho:
Fix: add _approx_equal_price_rounding to allow price in accepable range

Amount difference can be caused by rounding errors in the unit price.
Schwab rounds down the price to 4 decimal places so that the error in amount can be more than $0.01.
Fox example:
500 shares of FOO sold at $100.00016 with $1.23 fees.
"01/01/2024,"Sell","FOO","FOO","500","$100.0001","$1.23","$49,998.85"
calculated_amount = 500 * 100.0001 - 1.23 = 49998.82
amount_on_record = 49998.85 vs calculated_amount = 49998.82

In this PR change, the acceptable range of unit price is within $0.0001 when compare price on record and from retro calculation.

### DIFF
--- a/cgt_calc/main.py
+++ b/cgt_calc/main.py
@@ -31,6 +31,7 @@ from .initial_prices import InitialPrices
 from .model import (
     ActionType,
     BrokerTransaction,
+    CalcuationType,
     CalculationEntry,
     CalculationLog,
     CapitalGainsReport,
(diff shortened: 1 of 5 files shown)


cgt_calc/main.py
```diff
@@ -57,6 +58,55 @@ def get_amount_or_fail(transaction: BrokerTransaction) -> Decimal:
...
+            "Price calculated_price %.6f vs price_on_record %s in an acceptable range",
+            calculated_price,
+            price_on_record,
+        )
+        return True
+    LOGGER.debug(
+        "Price error in calculated_price %.6f vs price_on_record %s",
+        calculated_price,
+        price_on_record,
+    )
```
vmartinv: This logs the same information as the lines below, can you please remove it?

hermanho: Updated

PR #572, opened 2024-11-29
Updating T212 parser

aldrinpinto:


KapJI:
Please remove publish workflow. There is already one.

aldrinpinto:
Updated the parser and removed the public workflow

KapJI:
Thanks!

Issue #574, opened 2024-12-06
schwab_equity_award_json parser can introduce rounding errors in share quantity


m01:
The schwab_equity_award_json parser can introduce rounding errors when selling an integer number of shares, which can trigger asserts later.

The workaround in https://github.com/KapJI/capital-gains-calculator/blob/main/cgt_calc/parsers/schwab_equity_award_json.py#L232-L282 for dealing with missing decimal numbers after the decimal point in some early Schwab exports can introduce rounding errors in the quantity of shares being sold, which can later trigger asserts (e.g. the one that checks if you're selling more shares than you have).

The problematic line is https://github.com/KapJI/capital-gains-calculator/blob/main/cgt_calc/parsers/schwab_equity_award_json.py#L282.

I'll try to come up with a repro and will update this bug accordingly. (EDIT: done, see #575)

I think a simple fix is to check if you can reproduce the sale total `$` number from the transaction using the per share price, quantity, fees and rounding.

PR #575, opened 2024-12-14
schwab_equity_award_json.py parser: avoid unnecessary quantity recalculation.

m01:
This can introduce rounding errors.

This fixes https://github.com/KapJI/capital-gains-calculator/issues/574.

Basically, there's some code in `cgt_calc/parsers/schwab_equity_award_json.py`, which tries to work around some issues in data exported from Schwab:
```
            # Schwab's data export sometimes lacks decimals on Sales
            # quantities, in which case we infer it from number of shares in
            # sub-transactions, or failing that from the amount and salePrice.
```
This codepath is triggered unconditionally when an integer number of shares is sold - even when there isn't an issue with the exported data. This change makes that codepath conditional on there actually being an issue with the exported data.

Without the change in `cgt_calc/parsers/schwab_equity_award_json.py`, the test triggers the following error:
```
FAILED tests/test_schwab_equity_award_json.py::test_schwab_transaction_v2_rounding - AssertionError: assert Decimal('91.00001463507441935342241215') == Decimal('91')
```

The test data is fictional and purely there to reproduce the error, but it is plausible based on data I've seen and I have seen the rounding errors happen in real data.

Note that because quantities matter for S104 pool calculations this may introduce deltas in outputs between
calculations before and after this change. However, without this change you may run into asserts, depending
on what kinds of trades/gifts you make.

This contribution has been developed in my spare time. Disclaimer: I'm not a financial advisor/accountant.

### DIFF
--- a/cgt_calc/parsers/schwab_equity_award_json.py
+++ b/cgt_calc/parsers/schwab_equity_award_json.py
@@ -196,7 +196,8 @@ class SchwabTransaction(BrokerTransaction):
                 f"Schwab Equity Award JSON only supports GOOG stock but found {symbol}",
             )
         quantity = _decimal_from_number_or_str(row, names.quantity)
-        amount = _decimal_from_number_or_str(row, names.amount)
+        initially_parsed_amount = _decimal_from_number_or_str(row, names.amount)
+        amount = initially_parsed_amount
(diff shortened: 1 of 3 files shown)


KapJI:
Looks good to me, thanks for adding tests!

PR #577, opened 2024-12-23
Add calculating dividends gains.

nihn:
Cover the double taxation rules and tax treaties for dividends payments - the country of origin for a dividend is determinated by looking at its currency, this is not perfect but works most of the time (will not work for example for all EU countries using the same currency - this will need to be handled later on separately).

I've added a simple validation check which compares the tax at source with a one expected for the detected country which should remove (but not fix) some of the invalid matches.

The code will generate a similar pdf report as for capital gains. Right now only treaty with US and Poland are added to the calculations.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,5 +7,6 @@ __pycache__
 env
 dist
 calculations.pdf
+dividends.pdf
 exchange_rates.csv
 spin_offs.csv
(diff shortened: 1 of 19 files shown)


vmartinv:
I actually needed this one! Can you take a look why is it failing now?

ivendor:
I've reworked the change as per private discussions with @vmartinv  in the new PR linked here

nihn:
Closing as @ivendor forked it and I do not have time to finish anyway.

PR #587, opened 2025-01-27
Add support for annual HMRC exchange rates

calmarj:
Download annual exchange rates from HMRC and option to use them instead of monthly.

Had to use library to map country name to currencyCode as different years have completely different formats of the file.

Tested by running: 

`cgt-calc --year 2023 --schwab files/schwab.csv --schwab-award files/award.csv --exchange-rate-type annual`


KapJI:
Why do you want to use yearly rates instead of monthly?

PR #592, opened 2025-04-09
parsers/raw - parse single file and whole folder

donpiekarz:
Adding support for RAW parser to parse single files or whole folders, similar to other parsers.


KapJI:
Thanks! Can you fix typing as well?

PR #594, opened 2025-07-26
Update HMRC trade tariff API usage

Snehal-Reddy:
https://www.trade-tariff.service.gov.uk/uk/api/v2/.. now returns 404, and according to the links [here](https://www.trade-tariff.service.gov.uk/exchange_rates/monthly), "https://www.trade-tariff.service.gov.uk/uk/api/" seems to be the correct one now

### DIFF
--- a/cgt_calc/currency_converter.py
+++ b/cgt_calc/currency_converter.py
@@ -90,7 +90,7 @@ class CurrencyConverter:
         else:
             month_str = date.strftime("%Y-%m")
             url = (
-                "https://www.trade-tariff.service.gov.uk/api/v2/"
+                "https://www.trade-tariff.service.gov.uk/api/"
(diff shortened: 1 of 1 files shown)


cgt_calc/currency_converter.py
```diff
@@ -90,7 +90,7 @@ def _query_hmrc_api(self, date: datetime.date) -> None:
         else:
             month_str = date.strftime("%Y-%m")
             url = (
-                "https://www.trade-tariff.service.gov.uk/api/v2/"
+                "https://www.trade-tariff.service.gov.uk/api/"
```
KhaledHelmy: The URL is also missing a `/uk/` before `api`, so it should be: "https://www.trade-tariff.service.gov.uk/uk/api/".

Snehal-Reddy: My bad, fixed.

PR #598, opened 2025-09-16
Add Vanguard Parser

ivendor:
On Vanguard you can request an Excel report for all transactions for all your accounts.
The parser assumes that you get a report extract all the transaction data from the appropriate Excel tab (in case you have multple accounts like ISA and GA) into a single CSV and run it.

I've added the handling for all the cases I had in my spreadsheet

Added example test case.


ivendor:
Added the documentation in README.md

PR #599, opened 2025-09-20
Add interest and dividends breakdown per broker/symbol

ivendor:
Interests should be reported in different sections of self-assesment depending if they are UK or foreign.
Dividends in addition can be reported as Interest or Dividends based on the type of Fund (Bond vs Shares).
To ease out these I'm adding breakdown per broker of interest received and per symbol of dividends.

Changed tests appropriately


ivendor:
rebase recent changes

PR #564, opened 2024-09-28
Implement Freetrade parser

podtserkovskiy:
Supported events
- Sell
- Buy
- Top up
- Withdrawal
- Interest on cash
- Dividends
- Free share

Freetrade doesn't provide stock splits events, I added a hacky support for few stocks borrowing code from another parser

PR #600, opened 2025-09-21
Add FreeTrade parser

jeremychan:
Add parser for FreeTrade which is another broker similar to Trading212


vmartinv:
Make sure your implementation works as well as the pending one in #564  (I see some differences)

vmartinv:
Setting PR to draft until errors are fixed

Issue #573, opened 2024-12-01
`Total disposal proceeds` excludes fees


m01:
TL;DR: it looks to me like fees are removed from the disposal proceeds instead of added to allowable costs. I would've thought they should be added to allowable costs.

Reading the `Capital gains summary notes (2024)` from http://www.gov.uk/government/publications/self-assessment-capital-gains-summary-sa108 (see quote below), I get the impression that the `Total disposal proceeds` shown in the report should be the `num_shares * price` (or better, `amount + fees`) before the `fees` are taken off, and the transaction-specific fees should be part of the costs. The calculator seems to parse the `amount` as the `amount` after fees have been taken off (see e.g. https://github.com/KapJI/capital-gains-calculator/blob/main/cgt_calc/parsers/mssb.py#L161-L162) and then sum those to calculate the `Total disposal proceeds` (https://github.com/KapJI/capital-gains-calculator/blob/main/cgt_calc/main.py#L307).

> Sale or disposal proceeds
> Proceeds may include:
> * cash, payable now or in the future, or anything that can be turned into cash, unless it’s taxable as income
> * the market value of an asset you were given in exchange for the asset you disposed of
> * the value of a right to receive future payments where you do not know the amount of the payment at the time you sold it
>
> Sometimes you need to use the market value of the asset instead of the sale or purchase price.

> Allowable costs
> For Capital Gains Tax purposes, you deduct your costs to work out the gain or loss on the sale or disposal of an asset. Only some costs are allowable. These include:
> * the price paid to buy the asset
> * the costs of any improvements made to your asset – but they must be reflected in the asset when you dispose of it
> * incidental costs of acquiring or disposing of the asset, such as Stamp Duty or Stamp Duty Land Tax
>
> ...

What do others think? Did I misunderstand something?

vmartinv:
I believe it should include the fees. (which are part of the allowable costs too)

Sources:
- CG15250 https://www.gov.uk/hmrc-internal-manuals/capital-gains-manual/cg15250
- TCGA92/S38 https://www.legislation.gov.uk/ukpga/1992/12/section/38

vmartinv:
Fixed in #601 !

PR #601, opened 2025-09-21
Consider sell fees as an allowable cost

vmartinv:
As already proposed in #573, the fees incurred when you sell should be considered an allowable cost. Up until now it was just discounted from the disposal proceeds. Coincidentally both these methods lead to the same chargeable capital gain, see below:

### Current calculation (incorrect)
```
allowable_costs = buy_quantity * buy_price + buy_fees
disposal_proceeds = sell_quantity * sell_price - sell_fees
capital_gains = disposal_proceeds - allowable_costs
```

### Proposed calculation (correct)
```
allowable_costs = buy_quantity * buy_price  + buy_fees + sell_fees
disposal_proceeds = sell_quantity * sell_price
capital_gains = disposal_proceeds - allowable_costs
```

Besides the evidence provided in #573, you can see [this example from HMRC](https://assets.publishing.service.gov.uk/government/uploads/system/uploads/attachment_data/file/972646/HS284_Example_3_2021.pdf) for a simple case where the sell fees are clearly considered a cost.

# Implementation

I was more strict with the variable naming to avoid further confusions. To be clear:
- **amount** is the actual money received or sent, should be the difference on the cash balance of the account before and after the transaction.
- **disposal_proceeds** is the total money received for the shares, including what was spent in fees, i.e. `amount+fees`

# Testing

You will notice that no test had changes in the reported capital gains as expected. Only disposal proceeds and allowable costs had to be adjusted, which is a relief. No under or over-payment has been done.


KapJI (approved):
Thank you!

PR #603, opened 2025-09-24
Add support for dividends and their taxes in CGT report

ivendor:
This is a rework of https://github.com/KapJI/capital-gains-calculator/pull/577

The main difference is that in this PR we have a single merged PDF with all the events and proper distinction between funds/shares yields taxed as dividends and fund yields taxed as interests (bond funds) in UK.

As in the original PR we take into account double taxation treaties to calculate dividend taxes.


ivendor:
In the last push I've properly split the report for Capital Gains and Interest/Dividend events

cgt_calc/model.py
```diff
@@ -42,6 +55,31 @@ def __add__(self, transaction: HmrcTransactionData) -> HmrcTransactionData:
...
+        """Add two amounts."""
+        assert (
+            self.currency or not self.amount
+        ), f"Invalid foreign currency amount {self}"
+        assert (
+            amount.currency or not amount.amount
+        ), f"Invalid foreign currency amount {amount}"
+        result = ForeignCurrencyAmount(
+            amount=self.amount + amount.amount,
+            currency=self.currency or amount.currency,
```
vmartinv, line 78: add an assert that `self.currency==amount.currency` before this

cgt_calc/main.py
```diff
@@ -390,19 +410,50 @@ def convert_to_hmrc_transactions(
...
                 new_balance += amount
                 if self.date_in_tax_year(transaction.date):
-                    interest += self.converter.to_gbp_for(amount, transaction)
+                    if transaction.currency == COUNTRY_CURRENCY:
+                        interest_uk += amount
+                        rule_prefix = "interestUK"
+                    else:
+                        amount = self.converter.to_gbp_for(amount, transaction)
+                        interest_foreign += amount
+                        rule_prefix = "interestForeign"
```
vmartinv: let's not overwrite the variables
```suggestion
                    gbp_amount = self.converter.to_gbp_for(amount, transaction)
                    if transaction.currency == COUNTRY_CURRENCY:
                        interest_uk += gbp_amount
                        rule_prefix = "interestUK"
                    else:
                        interest_foreign += gbp_amount
                        rule_prefix = "interestForeign"
```

cgt_calc/main.py
```diff
@@ -427,16 +478,24 @@ def convert_to_hmrc_transactions(
                 )
                 raise CalculationError(msg)
             balance[(transaction.broker, transaction.currency)] = new_balance
+
+        sum_dividends, sum_dividends_tax, sum_div_interests = self.process_dividends(
+            dividends, dividends_tax
+        )
```
vmartinv: I feel this is breaking a bit the design of the class. Currently we are using two stages:
Stage 1 "**convert_to_hmrc_transactions**". Convert broker transactions into some standardized data struct, like the list of transactions `acquisition_list`/`disposal_list` and using `add_acquisition`/`add_disposal` functions.
Stage 2 "**calculate_capital_gain()**" Compute final results by iterating each day and calling `process_acquisition`/`process_disposal` functions on the data of the previous stage.

I think you can easily follow this pattern for your changes. We don't need to do everything in the first pass now that the logic is becoming more complex.
Stage 1: Populate the `dividend_list`, `interests_list` (or anything you need) using `add_dividend`.
Stage 2: Call `process_dividend` within the day iteration and accumulate the results.

Stage two could either be separate or within `calculate_capital_gain()`

cgt_calc/main.py
```diff
@@ -135,11 +145,16 @@ def __init__(
...
         self.calc_unrealized_gains = calc_unrealized_gains
+        self.interest_fund_tickers = interest_fund_tickers
+        self.total_uk_interest = Decimal(0)
+        self.total_foreign_interest = Decimal(0)
 
         self.acquisition_list: HmrcTransactionLog = {}
         self.disposal_list: HmrcTransactionLog = {}
         self.bnb_list: HmrcTransactionLog = {}
 
+        self.calculation_log_yields: CalculationLog = defaultdict(dict)
```
vmartinv, line 170: add a comment saying this is for interest and dividends, "yields" is not completely clear to me

cgt_calc/resources/template.tex.j2
```diff
@@ -119,12 +123,62 @@
...
 \subsection*{Number of acquisitions: \VAR{ acquisition_count.value }}
 \subsection*{Number of disposals: \VAR{ disposal_count.value }}
+\subsection*{Number of dividend distributions: \VAR{ dividend_count.value }}
+\subsection*{Number of interest transactions: \VAR{ interest_count.value }}
 \subsection*{Total disposal proceeds: £\VAR{ "{:,}".format(report.disposal_proceeds) }}
 \subsection*{Total capital gain before loss: £\VAR{ "{:,}".format(total_gain.value) }}
 \subsection*{Total capital loss: £\VAR{ "{:,}".format(total_loss.value) }}
 \subsection*{Total capital gain after loss: £\VAR{ "{:,}".format(total_gain.value - total_loss.value) }}
-
+\subsection*{Total amount of dividends proceeds: £\VAR{ "{:,}".format(round_decimal(report.total_dividends_amount(), 2)) }}
```
vmartinv, line 181: can you also add the dividend allowance used (the one defined by the tax year)

cgt_calc/resources/template.tex.j2
```diff
@@ -119,12 +123,62 @@
...
+
+    \BLOCK{ endfor }
+\BLOCK{ endif }
+
+\clearpage
 \section*{Overall}
 \subsection*{Number of acquisitions: \VAR{ acquisition_count.value }}
 \subsection*{Number of disposals: \VAR{ disposal_count.value }}
+\subsection*{Number of dividend distributions: \VAR{ dividend_count.value }}
+\subsection*{Number of interest transactions: \VAR{ interest_count.value }}
```
vmartinv: move this down with the rest of the dividend/interest numbers

cgt_calc/main.py
ivendor: The need of the 2 passes is strictly because the need of calculating all the complex rules like bed and breakfasting which cannot be calculated logically within a single pass. 

Dividends and interests don't require this complexity (requiring side effects on the class, instead of pure functional approach which is simpler to understand and maintain) and adding more model classes (as we can't re-use HMRC transaction it is) and thus more code to strictly follow a model that doesn't apply to the specific functionality feels wrong to me, from maintenance and simplicity point of view.

vmartinv: You are already doing two passes and added an additional class (Dividend). Anyways, I think that's fine. What I'm saying specifically is that the code I highlighted can be moved to the second pass instead. My suggestion is not to add more complexity but rather organise what you already have.

cgt_calc/main.py
```diff
@@ -427,16 +481,21 @@ def convert_to_hmrc_transactions(
...
         for (broker, currency), amount in balance.items():
             print(f"  {broker}: {round_decimal(amount, 2)} ({currency})")
-        print(f"Dividends: £{round_decimal(dividends, 2)}")
-        print(f"Dividend taxes: £{round_decimal(-dividends_tax, 2)}")
-        print(f"Interest: £{round_decimal(interest, 2)}")
+        print("Dividends:")
+        for (symbol, currency), amount in dividends.items():
+            tax = dividends_tax[(symbol, currency)]
+            tax_str = f"{tax} taxed at source " if tax < 0 else ""
+            print(f"  {symbol}: {round_decimal(amount, 2)} {tax_str}({currency})")
```
vmartinv:
```suggestion
            tax_str = f", excluding {-tax} taxed at source" if tax < 0 else ""
            print(f"  {symbol}: {round_decimal(amount, 2)}{tax_str} ({currency})")
```

cgt_calc/model.py
```diff
@@ -284,4 +397,18 @@ def __str__(self) -> str:
                     " Take a look at the symbols with unknown unrealized gains above"
                     " and factor in their prices.\n"
                 )
+        out += (
```
vmartinv, line 400: Can you also print the `self.dividend_allowance` ? This way the user can verify the calculation done in `total_dividend_taxable_gain`

vmartinv (changes requested):
Thanks, this looks much better! Left some nits. Also, can you regenerate `calculations_example.pdf` to show the newly added sections? This file is referenced in the `README.md`.
One thing I'm not a fan of is that my own report for last year went from 8 to 34 pages because Trading212 gives me daily interest on the cash balance. Although I don't want to delay this PR for that, the PDF can be improved later. It looks like this (amounts cropped):
Screenshot 2025-09-28 at 17 58 37

ivendor:
Adding monthly group by for dividends (using last dividend date in the taxable month)

cgt_calc/main.py
```diff
@@ -390,19 +411,36 @@ def convert_to_hmrc_transactions(
...
+                amount = get_amount_or_fail(transaction)
+                symbol = transaction.symbol
+                currency = transaction.currency
+                if symbol is None:
+                    raise SymbolMissingError(transaction)
+                new_balance += amount
+                self.dividend_tax_list[(symbol, transaction.date)] += (
+                    ForeignCurrencyAmount(amount, currency)
+                )
+                dividends_tax[(symbol, currency)] += amount
```
vmartinv: only if within the tax year

cgt_calc/main.py
```diff
@@ -390,19 +411,36 @@ def convert_to_hmrc_transactions(
...
-                    dividends_tax += self.converter.to_gbp_for(amount, transaction)
             elif transaction.action is ActionType.INTEREST:
                 amount = get_amount_or_fail(transaction)
                 new_balance += amount
-                if self.date_in_tax_year(transaction.date):
-                    interest += self.converter.to_gbp_for(amount, transaction)
+                self.interest_list[(transaction.broker, transaction.date)] += (
+                    ForeignCurrencyAmount(amount, transaction.currency)
+                )
+                interests[(transaction.broker, transaction.currency)] += amount
```
vmartinv: only if within the tax year

cgt_calc/main.py
```diff
@@ -390,19 +411,36 @@ def convert_to_hmrc_transactions(
...
+                if symbol is None:
+                    raise SymbolMissingError(transaction)
                 new_balance += amount
-                if self.date_in_tax_year(transaction.date):
-                    dividends += self.converter.to_gbp_for(amount, transaction)
-            elif transaction.action in [ActionType.TAX, ActionType.ADJUSTMENT]:
+                self.dividend_list[(symbol, transaction.date)] += ForeignCurrencyAmount(
+                    amount, currency
+                )
+                dividends[(symbol, currency)] += amount
```
vmartinv: only if within the tax year

cgt_calc/main.py
```diff
@@ -427,16 +465,22 @@ def convert_to_hmrc_transactions(
...
-        print(f"Dividend taxes: £{round_decimal(-dividends_tax, 2)}")
-        print(f"Interest: £{round_decimal(interest, 2)}")
+        print("Dividends:")
+        for (symbol, currency), amount in dividends.items():
+            tax = dividends_tax[(symbol, currency)]
+            tax_str = f", excluding {-tax} taxed at source" if tax < 0 else ""
+            print(f"  {symbol}: {round_decimal(amount, 2)}{tax_str} ({currency})")
+        print("Interests:")
+        for (broker, currency), amount in interests.items():
+            print(f"  {broker}: {round_decimal(amount, 2)} ({currency})")
```
vmartinv:
```suggestion
        if interests:
          print("Interests:")
          for (broker, currency), amount in interests.items():
              print(f"  {broker}: {round_decimal(amount, 2)} ({currency})")
```

cgt_calc/resources/template.tex.j2
```diff
@@ -119,12 +123,65 @@
...
+\clearpage
+\section*{Overall - Capital Gains}
 \subsection*{Number of acquisitions: \VAR{ acquisition_count.value }}
 \subsection*{Number of disposals: \VAR{ disposal_count.value }}
 \subsection*{Total disposal proceeds: £\VAR{ "{:,}".format(report.disposal_proceeds) }}
 \subsection*{Total capital gain before loss: £\VAR{ "{:,}".format(total_gain.value) }}
 \subsection*{Total capital loss: £\VAR{ "{:,}".format(total_loss.value) }}
 \subsection*{Total capital gain after loss: £\VAR{ "{:,}".format(total_gain.value - total_loss.value) }}
-
+\vspace{1em}
```
vmartinv:
```suggestion
\vspace{2em}
```

cgt_calc/main.py
```diff
@@ -427,16 +465,22 @@ def convert_to_hmrc_transactions(
...
         for (broker, currency), amount in balance.items():
             print(f"  {broker}: {round_decimal(amount, 2)} ({currency})")
-        print(f"Dividends: £{round_decimal(dividends, 2)}")
-        print(f"Dividend taxes: £{round_decimal(-dividends_tax, 2)}")
-        print(f"Interest: £{round_decimal(interest, 2)}")
+        print("Dividends:")
+        for (symbol, currency), amount in dividends.items():
+            tax = dividends_tax[(symbol, currency)]
+            tax_str = f", excluding {-tax} taxed at source" if tax < 0 else ""
+            print(f"  {symbol}: {round_decimal(amount, 2)}{tax_str} ({currency})")
```
vmartinv:
```suggestion
        if dividends:
          print("Dividends:")
          for (symbol, currency), amount in dividends.items():
              tax = dividends_tax[(symbol, currency)]
              tax_str = f", excluding {-tax} taxed at source" if tax < 0 else ""
              print(f"  {symbol}: {round_decimal(amount, 2)}{tax_str} ({currency})")
```

vmartinv (changes requested):
Wow thanks, now mine is only 11 pages. Left some more comments.

vmartinv:
Code looks good to me now! Since this is a big change I leave it to @KapJI to give it a final look.

KapJI:
Looks great, thank you!

PR #605, opened 2025-10-04
Schwab Equity Award: support dividends and their taxes - takeover of #568

PierfrancescoSoffritti:
This PR is a takeover of #568. It fixes a typo in the tests date that made the test fail.


vmartinv:
Yes we added dividends and interests, also checkout the README on Schwab . I don’t think we should be using json anymore.

PR #609, opened 2025-10-05
Migrate from Poetry to uv

KapJI:
It's 2025 and it's time to migrate to uv here as well.

uv is another great tool by [Astral](https://docs.astral.sh/uv/). It's written in Rust and much faster than Poetry.
It also follows [PEP 517](https://peps.python.org/pep-0517/).

I also wrote `CONTRIBUTING.md` with guide how to use it.

@vmartinv @ivendor do you want to take a look? It's large but everything is split into isolated commits.


CONTRIBUTING.md
```diff
@@ -0,0 +1,102 @@
...
+cd capital-gains-calculator
+```
+
+### 3. Set up the environment
+
+```shell
+uv sync
+```
+
+This sets up the development environment with all dependencies installed.
```
ivendor: does it also create the virtual environment for you?  or does it install dependencies in the current active environment? maybe this can be specified better for people that don't know how uv works

KapJI: Yes, quite similar to poetry.

pyproject.toml
```diff
@@ -1,56 +1,67 @@
...
+]
 
-[build-system]
-requires = ["poetry-core>=1.0.0"]
-build-backend = "poetry.core.masonry.api"
+[tool.uv.build-backend]
+module-root = "."
 
 [tool.pylint.main]
 py-version = "3.9"
```
ivendor, line 67: it would be nice to update this as well I'm currently using python 3.13 on my laptop but up to you anyway

KapJI: I'm going to upgrade to 3.12 once this is merged.

ivendor (approved):

vmartinv (approved):
lgtm!

PR #610, opened 2025-10-05
Fix corner cases with ERI and polish

ivendor:
A few fixes for recent ERI changes, I discovered while trying to add support for more ETF ERI data.

1) There can be multiple ERI report within a single year for a fund, when the fund is closing. This means that B&B calculations needs to take into account multiple ERI events. I have adapted the models and the output for such use case.

2) In the ERI raw parsers from resources I skip any file starting from . to avoid .DS_Store stuff created by macos for example

3) I disabled the ability of tests to generate false ISIN translation in the output file as we use test data

4) I've refenerated the example file and made the example using B&B with ERI


cgt_calc/model.py
```diff
@@ -225,7 +225,7 @@ def __init__(
         bed_and_breakfast_date_index: datetime.date | None = None,
         spin_off: SpinOff | None = None,
         dividend: Dividend | None = None,
-        eri: ExcessReportedIncome | None = None,
+        eris: tuple[ExcessReportedIncome, ...] = (),
```
KapJI: Why not `list`?

ivendor: because default values of list cannot be empty list (mutable default arguments) and I didn't want to handle None, but I'm happy to accept suggestions

KapJI: I think list is semantically more correct for collections which store elements of the same type.

ivendor: Fair enough be ready for all the `or []`

cgt_calc/parsers/eri/raw.py
```diff
@@ -34,6 +35,8 @@ def __init__(self, header: list[str], row_raw: list[str], file: str):
         row = dict(zip(header, row_raw))
 
         isin = row["ISIN"]
+        if not is_isin(isin):
+            raise ParsingError(file, f"Not valid ISIN {isin}")
```
KapJI, line 39:
```suggestion
            raise ParsingError(file, f"Not valid ISIN: {isin}")
```

cgt_calc/main.py
```diff
@@ -443,17 +442,6 @@ def add_eri(
...
-                if (
-                    symbol in report_by_symbol
-                    and abs((report_date - transaction.date).days) < MIN_DAYS_IN_YEAR
-                ):
-                    raise InvalidTransactionError(
-                        transaction,
-                        "A reporting period within less than a year for this "
-                        f"ticker already exist at {report_date}",
-                    )
-
```
vmartinv, line 456: Maybe not remove this. We still need to check if we have two in the same day, and if they are within the same day whether the amount is different. It can be the case that the user attempts to add the same ERI twice, we should add a warning. If the amount is different then raise an error.

tests/test_vanguard.py
```diff
@@ -15,6 +15,8 @@ def test_run_with_vanguard_files() -> None:
         "tests/test_data/vanguard/report.csv",
         "--interest-fund-tickers",
         "FOO",
+        "--isin-translation-file",
+        "",
```
vmartinv: why this is needed? can you make it work without this? If anything we should raise an error because the user is not using the parameter correctly,

cgt_calc/model.py
```diff
@@ -83,15 +83,15 @@ class HmrcTransactionData:
     quantity: Decimal = Decimal(0)
     amount: Decimal = Decimal(0)
     fees: Decimal = Decimal(0)
-    eri: ExcessReportedIncome | None = None
+    eris: list[ExcessReportedIncome] = field(default_factory=list)
```
vmartinv, line 89: maybe add a comment on when this is used and why it's a list

vmartinv (changes requested):
This is an annoying corner case and I see no better way than transforming the eri field into a list, so let's go with it.

tests/test_vanguard.py
ivendor: This is to prevent test to pollute the isin translation files, as test have fake ISIN mappings from the trading212 transactions (or any other fake data in the test) and when you run them with pytest they're going to write in the same isin translation file you use for your normal transactions (if you're like me that runs the tool directly from the package folder for example).

Generally the principle is that tests should not pollute the filesystem when you run them, so the empty string prevent the writing of the file.

KapJI: You can just add it in `build_cmd`. Or even better, replace `build_cmd` with `subprocess.run` wrapper and set some env var like `CGT_TEST=1` for all integration tests. Then you can use it to disable logic which should be running for tests.

cgt_calc/const.py
```diff
@@ -4,12 +4,15 @@
...
 import datetime
 from decimal import Decimal
+import os
 from typing import Final
 
 from dateutil.relativedelta import relativedelta
 
 from .model import TaxTreaty
 
+CGT_TEST_MODE = "CGT_TEST_MODE" in os.environ
```
KapJI: It should be false if `CGT_TEST_MODE=0`.

PR #613, opened 2025-10-06
Upgrade Python to 3.12

KapJI:
Python 3.9 is EOL. Upgrade to Python 3.12. It should be available on all modern platforms and uv can install it if needed.


vmartinv (approved):
looks good as it is, I'd guess most of the `strict=False` can be removed though.

PR #614, opened 2025-10-06
Use zip(strict=True) for extra protection

KapJI:
Follow up to Python 3.12 upgrade.

@vmartinv as you added that Trading212 test csv. Do you know if this is a typo or Trading212 provided invalid CSV?

### DIFF
--- a/cgt_calc/parsers/sharesight.py
+++ b/cgt_calc/parsers/sharesight.py
@@ -10,7 +10,11 @@ from pathlib import Path
 from typing import Final
 
 from cgt_calc.const import TICKER_RENAMES
-from cgt_calc.exceptions import InvalidTransactionError, ParsingError
+from cgt_calc.exceptions import (
+    InvalidTransactionError,
+    ParsingError,
+    UnexpectedColumnCountError,
+)
 from cgt_calc.model import ActionType, BrokerTransaction
 
 STOCK_ACTIVITY_COMMENT_MARKER: Final[str] = "Stock Activity"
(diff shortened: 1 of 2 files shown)


cgt_calc/main.py
```diff
@@ -572,7 +572,7 @@ def convert_to_hmrc_transactions(
                     [
                         f"{trx}\nBalance after transaction={balance_after}"
                         for trx, balance_after in zip(
-                            transactions[: i + 1], balance_history, strict=False
+                            transactions[: i + 1], balance_history, strict=True
```
vmartinv, line 575: strict=True is the default behavior from Python 3.10, so we can just remove the parameter.
```suggestion
                            transactions[: i + 1], balance_history
```

tests/test_data/trading212/from_2020-09-11_to_2021-04-02.csv
```diff
@@ -1,7 +1,7 @@
 Action,Time,ISIN,Ticker,Name,No. of shares,Price / share,Currency (Price / share),Exchange rate,Result (GBP),Total (GBP),Withholding tax,Currency (Withholding tax),Charge amount (GBP),Transaction fee (GBP),Finra fee,Currency (Finra fee),Notes,ID
-Deposit,2020-06-24 04:06:06,,,,,,,,,9000.00,,,9000.00,,,"Transaction ID: xxxxxxx",123
+Deposit,2020-06-24 04:06:06,,,,,,,,,9000.00,,,9000.00,,,"Transaction ID: xxxxxxx",,123
```
vmartinv, line 2: I can confirm this was just a mistake I had when stripping out personal details from the csv. This change is correct.

cgt_calc/main.py
KapJI: It's not. It was added in Py3.10 but default it false: https://docs.python.org/3.10/library/functions.html#zip

vmartinv: okay in that case this is bettter then

vmartinv (approved):

PR #659, opened 2025-10-27
Skip bed-and-breakfast passes with zero quantity

carrithers:
## Summary
- Guard the bed-and-breakfast loop so it skips iterations when there’s no quantity left to match. Previously, the loop kept scanning the 30-day window even after the disposal was fully satisfied. When it hit the next acquisition, `available_quantity` dropped to zero, yet we still attempted `current_amount / current_quantity` where `current_quantity` was also zero—raising `decimal.DivisionUndefined`.
- Add a regression test (`test_bed_and-breakfast_zero_available_quantity_skip`) that mirrors the real-life pattern: a disposal immediately re-bought in full, followed by another acquisition inside the window. The old code crashed in that scenario; the new guard ensures it passes safely.

## Issue Details
A disposal that is completely matched by a prior acquisition within the 30-day window can still have additional acquisitions later in that same window. Without the guard, the loop treated those later rows as though they needed matching, despite the disposal being fully satisfied. That led to a 0 ÷ 0 division and aborted the calculation.

## Testing
- `uv run pytest tests/general/test_calc.py::test_bed_and_breakfast_zero_available_quantity_skip`
- `uv run pytest tests/general/test_calc.py`

### DIFF
--- a/cgt_calc/main.py
+++ b/cgt_calc/main.py
@@ -833,21 +833,18 @@ class CapitalGainsCalculator:
                         == 0
                     ):
                         continue
-                    available_quantity = min(
-                        disposal_quantity,
-                        acquisition.quantity
-                        - same_day_disposal.quantity
-                        - bnb_acquisition.quantity,
-                    )
-                    # Skip if no shares available for bed and breakfast
-                    if available_quantity == 0:
-                        continue
                     LOGGER.warning(
                         "Bed and breakfasting for %s. "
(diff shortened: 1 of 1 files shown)


cgt_calc/main.py
vmartinv, line 925:
```suggestion
                    # If we completely matched the current disposal,
                    # there's no need to keep looking for more B&D days
                    if disposal_quantity <= 0:
                        break
```

PR #660, opened 2025-10-27
Allow tiny rounding residue in bed-and-breakfast

carrithers:
## Summary
- Treat fractional rounding residue in the bed-and-breakfast logic as benign instead of aborting. Previously we asserted `round_decimal(current_amount, 23) == 0`; in real datasets the cumulative arithmetic can leave a microscopic remainder (e.g., `-2E-23`) even when the pool has been exhausted. We now allow any remainder smaller than `1E-20` and keep the safety check for larger discrepancies.
- Add a regression test (`test_bed_and_breakfast_rounding_residue`) that reproduces the scenario: a disposal, a matching re-buy, and a subsequent bed-and-breakfast entry that previously tripped the assertion because of the tiny residue.

## Issue Details
Repeated `Decimal` operations on fractional quantities can accumulate noise that lives within the default precision. When bed-and-breakfast adjustments drew the pool back to zero, that noise meant the amount column was “close to zero” but not binary-equal to zero. The assertion was meant to catch incorrect bookkeeping but wound up rejecting legitimate cases.

## Testing
- `uv run pytest tests/general/test_calc.py::test_bed_and_breakfast_rounding_residue`
- `uv run pytest tests/general/test_calc.py`

### DIFF
--- a/cgt_calc/main.py
+++ b/cgt_calc/main.py
@@ -893,8 +893,9 @@ class CapitalGainsCalculator:
                     current_quantity -= available_quantity
                     current_amount -= amount_delta
                     if current_quantity == 0:
-                        assert round_decimal(current_amount, 23) == 0, (
-                            f"current amount {current_amount}"
+                        # Allow for tiny rounding errors (up to 1E-20)
+                        assert abs(current_amount) < Decimal("1E-20"), (
+                            f"current amount {current_amount} is not close enough to zero"
                         )
                     add_to_list(
                         self.bnb_list,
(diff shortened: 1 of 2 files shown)


cgt_calc/main.py
```diff
@@ -893,8 +893,9 @@ def process_disposal(
                     current_quantity -= available_quantity
                     current_amount -= amount_delta
                     if current_quantity == 0:
-                        assert round_decimal(current_amount, 23) == 0, (
-                            f"current amount {current_amount}"
+                        # Allow for tiny rounding errors (up to 1E-20)
```
KapJI, line 896: Isn't it the same as `round_decimal(current_amount, 20)`?

carrithers: My apologies. I had another issue that i was testing at the same time and got the resolutions mixed up. Please feel free to either close this PR or revert this assertion and keep the test.

PR #662, opened 2025-11-01
Schwab json: Accept empty string values as meaning 0.

thibwk:
Schwab started to occasionally send an empty string value to mean 0, instead of 'null', which was breaking the parser.


vmartinv (changes requested):
As above needs more testing, and we should be more careful into what we allow to be empty (only fees can be empty for example)

vmartinv (changes requested):
Sorry, we still need more thorough testing. I.e. let's load a json with a buy transaction containing an empty amount and make sure it fails.

vmartinv:
merging to avoid users facing this problem

thibwk:
@vmartinv thanks for merging. FYI Schwab does provide "Deposit" transactions with a null Amount, so sadly we can reject that.

FYI I did just create a follow-up PR https://github.com/cgt-calc/capital-gains-calculator/pull/1038 to improve the test coverage.

PR #669, opened 2025-11-23
Fix interest calculations.

nihn:
Same broker can have interests in multiple currencies for the same date - that was causing a crash before, now is fixed. Also simplified the interest calculations by moving all the dates to the first day of the month (previously this was only done if there were double interests in a given month)

### DIFF
--- a/cgt_calc/main.py
+++ b/cgt_calc/main.py
@@ -1082,13 +1082,26 @@ class CapitalGainsCalculator:
         monthly_interests: dict[
             tuple[str, str, datetime.date], ForeignCurrencyAmount
         ] = defaultdict(ForeignCurrencyAmount)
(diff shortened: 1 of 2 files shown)


vmartinv:
Just a comment that we shouldn’t move the interest to the first of the month. I think we need to make sure we don’t move interest transactions to a tax year different to the one they actually happened. I think the current logic is complicated for that reason.

nihn:
> If I receive interest on the 2nd April, wouldn’t this update it to 15th April? That would change its tax year.

it won't because we first check if the date is within the given tax year

nihn:
ah, ok, you are right, I was only thinking about the beginning of the tax year - not the end, I'll restore previous logic

KapJI:
@vmartinv can you take another look?

vmartinv:
LGTM

PR #682, opened 2025-12-14
Add Journaled Shares action type.

amaksoft:
Maps Schwab's "Journaled Shares" action to ActionType.TRANSFER to handle account-to-account transfers.

Observed on latest data, Schwab uses "Journaled Shares" to record transfers between accounts. Without this mapping, the parser fails with "Unknown action: Journaled Shares".


vmartinv:
so both of these accounts are yours right? If so it might be tricky because we need to somehow keep track of the pool cost of the shares across the accounts (I'm guessing that the price shown in the transaction is the current value, not the calculated cost). Well if you load both accounts in the calculator it should work.
The current `TRANSFER` action was thought for cash transfers not shares. CGT is different if you transfer to yourself, to someone else or to your married partner so we need to be careful here.

vmartinv:
If you delete these transfers then why do you need this PR? I would rather fail because share transfers are not yet covering all cases.
Your set up is a bit strange because most people have separate accounts to fully use their partner's allowance and I would rather start by supporting that first.

PR #683, opened 2025-12-14
Adjust CUSIP bond prices automatically.

amaksoft:
Adds automatic price adjustment for CUSIP bond symbols, which use pricing per $100 face value.

Context:
CUSIP bonds (9-character identifiers starting with a digit) are priced differently from stocks. A bond priced at $99.1727 per $100 face value needs to be divided by 100 to get the actual price per $1.

Real-world example:
 - Buy $40,000 face value at $99.1727 per $100 
 - CSV shows: Price=$9917.27, Quantity=40000, Amount=-$3,966,908.00
 - Adjusted: Price=$99.1727 (per $1), Quantity=40000
 - Calculation: 40000 × $99.1727 = $3,966,908.00 ✓
 
 Without the adjustment above, the validation fails


cgt_calc/parsers/schwab_cusip_bonds.py
```diff
@@ -0,0 +1,157 @@
...
+
+
+def _is_cusip_symbol(symbol: str | None) -> bool:
+    """Check if symbol is a CUSIP (9 alphanumeric chars starting with digit).
+
+    CUSIP format: 9 alphanumeric characters, typically starting with a digit.
+    We use a simplified check (length + starts with digit) rather than full
+    validation since broker data is assumed to be well-formed.
+    """
+    return bool(symbol and len(symbol) == CUSIP_SYMBOL_LENGTH and symbol[0].isdigit())
```
vmartinv: just to make sure we don't apply this to other asset types, can we strengthen the check here? It should be possible to validate the check digit (checksum). We can also restrict 7th and 8th to the ones we know to work well.

What a CUSIP Looks Like
- First 6 characters (the “CUSIP‑6”) identify the issuer (e.g., Apple Inc.).
- 7th and 8th characters identify the specific issue/type of security (stock class, bond series, etc.).
- 9th character is a check digit calculated from the prior 8 characters to validate the code.
​

amaksoft: Added the check. I don't want to limit the issuers as people are free to buy whatever bonds they want, but we check whether they are priced per $100 face value before recalculating. I believe this should be safe enough?

vmartinv:
Thanks for working on this!

PR #684, opened 2025-12-14
Parse Schwab "as of" dates correctly.

amaksoft:
Adds support for Schwab's "MM/DD/YYYY as of MM/DD/YYYY" date format, using the date after "as of" as the actual transaction date.

Context:
Schwab uses this format when the recorded date differs from the actual transaction date due to settlement delays, holidays, or timing issues. Using the wrong date causes transactions to be processed out of chronological order, leading to incorrect capital gains calculations.

Real-world example (RSU settlement):
- RSU vests on 08/15/2023
- Schwab records as "08/18/2023 as of 08/15/2023" (settlement delay)
- Tax withholding sell on 08/16/2023

Without this fix:
 - RSU grant would use 08/18 (wrong!)
 - Tax sell on 08/16 would sort BEFORE RSU grant
 - Error: "sell before acquire"

With this fix:
 - RSU grant correctly uses 08/15 (actual vest date)
 - Transactions sort correctly: Grant (08/15) → Sell (08/16)
 - Capital gains calculated accurately

### DIFF
--- a/cgt_calc/main.py
+++ b/cgt_calc/main.py
@@ -177,6 +177,11 @@ class CapitalGainsCalculator:
         self.disposal_list: HmrcTransactionLog = {}
         self.bnb_list: HmrcTransactionLog = {}
         self.split_list: dict[tuple[str, datetime.date], Decimal] = {}
+        self.vest_date_matched: dict[datetime.date, dict[str, Decimal]] = {}
+        # Track vest_date for acquisitions (date, symbol) -> vest_date
+        self.acquisition_vest_dates: dict[tuple[datetime.date, str], datetime.date] = {}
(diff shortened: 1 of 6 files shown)


vmartinv:
This is just reversing #667 , please check it out so we can find a better solution.

vmartinv:
It shouldn’t match when disposal_date < settlement_date , unless is a tax sell

amaksoft:
How do you propose to identify the "tax sell". It's not like they are labelled

Here I was going through all possible scenarios and trying to cover all the edge cases.

vmartinv:
Yeah it’s unclear what you’re trying to fix here

PR #686, opened 2025-12-14
Filter out cancelled Buy transactions for Schwab.

amaksoft:
Schwab-specific feature: When a purchase is cancelled, Schwab reports both the original Buy and a "Cancel Buy" transaction. Both need to be filtered out to avoid incorrect capital gains calculations.

This follows the same pattern as Cash Merger + Cash Merger Adj, where both transaction types map to the same ActionType, and the filtering logic uses raw_action to distinguish between them.

This is implemented entirely within schwab.py parser as it's a Schwab-specific CSV reporting quirk, not a general transaction type.


vmartinv (approved):
LGTM, wondering whether we will also need "Cancel sell" support.

PR #687, opened 2025-12-14
Add Full Redemption transaction support.

amaksoft:
Add support for Full Redemption corporate actions where bonds are redeemed by the issuer before (or at) maturity.

Without this the tool fails with "Unknown action: Full Redemption" error

Context:
Full Redemption follows a similar pattern to Cash Merger but with reversed transaction order. Schwab splits the transaction across two rows that must be combined to calculate the redemption price.

Also refactors existing Cash Merger logic to use the same unified pairing approach, reducing code duplication and ensuring consistent handling of similar corporate action patterns.


vmartinv:
this one looks good too, I don't see any major issues.

PR #696, opened 2025-12-25
refactor: encapsulate broker-specific logic into self-contained classes

vmartinv:
# Problem
Each parser's logic is not encapsulated in a single file, currently to add a parser you need edit several files: `args_parser` (to register the input args), `main` (to read the args), `parsers/__init__`  (to call your parser) and the parser code itself. This is hard for someone not familiar with the code.

# Solution
Refactor all parsers and create a `BrokerRegistry` class. With this change adding a parser is as easy as adding the parser code and add the class to the `BrokerRegistry`. The base classes supplied allow maintaining cohesion on the command line arguments and reduce redundancy. This should make reading and adding new parsers much simpler.

# Code review

Most of the PR is moving code around. Some parsers like ERIRaw received more changes to adapt to the new interface, but it was also the hardest to understand. It's possible to reduce redundancy even more but I will leave that for a later PR.

### DIFF
--- a/cgt_calc/current_price_fetcher.py
+++ b/cgt_calc/current_price_fetcher.py
@@ -47,7 +47,6 @@ class CurrentPriceFetcher:
             return self.historical_prices_data[symbol][date]
 
         prices = yf.Ticker(symbol).history(
-            period="1d",
             interval="1d",
             start=date.strftime("%Y-%m-%d"),
             end=(date + datetime.timedelta(days=1)).strftime("%Y-%m-%d"),
--- a/cgt_calc/main.py
+++ b/cgt_calc/main.py
@@ -63,7 +63,7 @@ from .parsers.broker_registry import BrokerRegistry
 from .setup_logging import setup_logging
 from .spin_off_handler import SpinOffHandler
 from .transaction_log import add_to_list, has_key
-from .util import approx_equal, round_decimal
+from .util import approx_equal, normalize_amount, round_decimal
 
 if TYPE_CHECKING:
     import argparse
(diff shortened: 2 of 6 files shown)


vmartinv:
@KapJI any opinions on this? I want to add a new parser afterwards (Interactive Brokers)

cgt_calc/parsers/broker_registry.py
```diff
@@ -0,0 +1,146 @@
...
+        """Register arguments for all brokers."""
+        for broker_class in BrokerRegistry._BROKERS:
+            broker_class.register_arguments(broker_group)
+
+        # ERI Raw is not a broker but is close enough to one to be here
+        ERIRawParser.register_arguments(broker_group)
+
+        BrokerRegistry._register_deprecated_arguments(broker_group)
+
+    @staticmethod
```
KapJI: Why these are here? Can they be moved to corresponding classes?

cgt_calc/parsers/broker_registry.py
```diff
@@ -0,0 +1,146 @@
...
+
+    from .base_parsers import BaseParser
+
+LOGGER = logging.getLogger(__name__)
+
+
+class BrokerRegistry:
+    """Registry for all broker parsers."""
+
+    _BROKERS: ClassVar[list[type[BaseParser]]] = [
```
KapJI, line 31: What do think about going a step further and have some class annotation which adds them to registry? Then single `from cgt_calc.parsers import *` should cover all of them.

vmartinv: The issue with this is we still need to import each parser individually in the `parsers/__init__.py`. There's a way to import all modules using pkgutils but it's rather hacky.
IMO I think is easier to understand if they just add new parsers to the BrokerRegistry list as it is now.

cgt_calc/parsers/broker_registry.py
vmartinv: Moved

KapJI (approved):

PR #701, opened 2025-12-30
Don't crash while determining version.

m01:
If the Distribution Package version cannot be determined, cgt-calc currently errors out with a `importlib.metadata.PackageNotFoundError`. This means that e.g. running `pytests` by hand fails, which seems unnecessarily restrictive.

This commit makes the version string fall back to "unknown version" in those cases, which should make development easier while also prompting the right questions if users raise an issue.

This contribution has been developed in my spare time.

### DIFF
--- a/cgt_calc/args_parser.py
+++ b/cgt_calc/args_parser.py
@@ -402,10 +402,14 @@ Environment variables:
         action="help",
         help="show this help message and exit",
     )
+    try:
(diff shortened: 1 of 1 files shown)


m01:
Arch Linux, Python 3.13.11, pytest 8.4.2.

I'm just running `pytest` instead of `uv run pytest` though.

KapJI:
You need to run test from uv environment and this should work then.

PR #704, opened 2026-01-11
Handle Schwab interest withholding taxes

lgyanf:
This PR fixes issue #705 

## Summary
  - Treat Schwab `NRA Tax Adj` rows without symbols and with `SCHWAB1 INT` as account-level interest tax.
  - Report interest tax totals as positive values and surface them in text/PDF outputs.
  - Add regression tests for interest tax handling and update expected outputs.

## Testing
  - `uv run pytest`
  - `uv run pytest tests/general/test_calc.py::test_interest_tax_totals_are_positive`
  - `uv run pytest tests/schwab/test_schwab_interest_tax.py`


vmartinv:
After merging, looks like the new test needs fixing

KapJI:
Rebased onto current main and pushed two commits (thanks for allowing maintainer edits):

- Your change ported onto the new parser architecture: reclassification now lives in the reworked `SchwabTransaction.__init__`, the test uses `SchwabParser.load_from_file`, `ActionType.INTEREST_TAX = 20` / `RuleType.INTEREST_TAX = 10` (19/9 were taken by RENAME in the meantime).
- Review follow-ups: monthly grouping extracted into a shared `_group_by_month` helper (also fixes month comparison across years — a UK tax year contains two Aprils), `total_foreign_interest_tax` renamed to `total_interest_tax` since the template already labels GBP entries as UK, and a test pinning that symbol-bearing `NRA Tax Adj` rows remain dividend tax.

PR #755, opened 2026-03-09
Feature/add transactions foreign currency ibkr

bogdanutanu:
Hi, first of all many thanks for your efforts developing this tool, it is invaluable for CGT calculations especially when holding accumulating funds.

After using it with my own transactions, I found out it wasn't supporting transactions in currencies other than GBP for Interactive Brokers. I'm proposing these changes to support parsing IBKR files which may have columns for exchange rate and foreign currency, converting the buy and sell prices into GBP, so all calculations are done in GBP as per HMRC rules.

I still have to run it with the `--no-balance-check` flags, due to how Interactive Brokers accounts for currency conversions. They include some extra transactions to account for adjustments when converting currencies, I still haven't got my head around those.

Let me know what you think!

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -51,10 +51,10 @@ jobs:
         uses: actions/checkout@v6
 
       - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v3
+        uses: docker/setup-buildx-action@v4
 
       - name: Build image
-        uses: docker/build-push-action@v6
+        uses: docker/build-push-action@v7
         with:
           context: .
           tags: local/cgt-calc:test
--- a/.github/workflows/manage-labels.yml
+++ b/.github/workflows/manage-labels.yml
@@ -14,4 +14,4 @@ jobs:
         uses: actions/checkout@v6
 
       - name: Run Labeler
-        uses: crazy-max/ghaction-github-labeler@v5.3.0
+        uses: crazy-max/ghaction-github-labeler@v6.0.0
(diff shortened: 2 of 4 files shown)


cgt_calc/model.py
```diff
@@ -179,6 +179,7 @@ class BrokerTransaction:
     currency: str
     broker: str
     isin: str | None = None
+    exchange_rate: Decimal | None = None
```
vmartinv: why are you adding this? I don't see any downstream usage of it and it is rather confusing if we are already doing the conversion during the parsing stage

cgt_calc/resources/eri/vanguard_eri.csv
```diff
@@ -1,4 +1,6 @@
 ISIN,Fund Reporting Period End Date,Currency,Excess of reporting income over distribution
+IE00BK5BQT80,30/06/2025,USD,2.2479
+IE00BG47KH54,30/06/2025,EUR,0.7740
```
vmartinv: let's keep this PR limited to improving IBKR. I think for ERI's is better if we add everything at once by parsing rather than review every 2 or 3.

cgt_calc/resources/eri/blackrock_eri.csv
```diff
@@ -0,0 +1,5 @@
+ISIN,Fund Reporting Period End Date,Currency,Excess of reporting income over distribution
+IE00BDBRDM35,30/06/2023,USD,0.1002
+IE00BDBRDM35,30/06/2024,USD,0.1373
+IE00B441G979,30/11/2022,EUR,0.8373
```
vmartinv: same thing, let's not add numbers without a proper parser. Until we have it you can keep these values locally.

PR #757, opened 2026-03-12
Fix Vanguard parsing for funds without symbols and fractional share quantities

newwaylw:
Vanguard parser improvements (cgt_calc/parsers/vanguard.py):
- Updated BOUGHT_RE and SOLD_RE regexes to handle funds that don't have a ticker symbol in parentheses (e.g. "Bought 6.77 Emerging Markets Stock Index Fund - Accumulation"). Previously, a parenthesized symbol like (FOO) was required. Now the full fund name is used as a fallback when no (SYMBOL) is present.
- Added support for fractional shares starting with a decimal point (e.g. .9232).
- Added "Cash transfer" and case-insensitive "Account fee" to the transfer regex.

Tests (tests/vanguard/test_vanguard.py):
- Added test_read_vanguard_missing_symbol — verifies the full fund name is used when no parenthesized symbol exists.
- Added test_read_vanguard_fractional_share — verifies parsing of quantities like .2.

Test data (tests/vanguard/data/):
- Added two new rows to report.csv for funds without ticker symbols.
- Updated expected_output.txt to reflect the new transactions and adjusted balance.

### DIFF
--- a/cgt_calc/parsers/vanguard.py
+++ b/cgt_calc/parsers/vanguard.py
@@ -30,11 +30,11 @@ class VanguardColumn(StrEnum):
 
 COLUMNS: Final[list[str]] = [column.value for column in VanguardColumn]
 
-BOUGHT_RE = re.compile(r"^Bought (\d*[,]?\d*) .*\((.*)\)$")
-SOLD_RE = re.compile(r"^Sold (\d*[,]?\d*) .*\((.*)\)$")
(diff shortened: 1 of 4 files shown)


newwaylw:
> “Account fee” should be a fee, not a transfer.

This I don’t know, I have not changed “Account fee” type. @ivendor can you confirm?

ivendor:
So the change looks good to me, thank you for doing it. About the FEEs, I always interepreted FEE as transactions FEE not accounts fees. I.e. the vanguard fees are for holding your accounts not on transaction so always treated them as simple negative cash transfer. Please feel free to correct me, but in general this is orthogonal to this PR

PR #760, opened 2026-03-15
Ease out updating of test exchange_rates when ERI or new test cases are added

ivendor:
When introducing new ERI values or new broker test case are added we have to now manually append entries into exchange rates. To workaround this problem, I've forked the test mode in 2 test modes:
- TEST MODE when the pytest is run outside of pre-commit hook which auto append entries to the test mocks supporting parallel execution of tests via file locks
- TEST STRICT MODE when the pytest is run inside the pre-commit hook which beahves like TEST MODE before this change.

Finally regenerated the CSV file using TEST MODE


cgt_calc/currency_converter.py
```diff
@@ -39,10 +40,16 @@ def __init__(
         """Load data from exchange_rates_file and optionally from initial_data."""
         self.exchange_rates_file = exchange_rates_file
         read_data = self._read_exchange_rates_file(exchange_rates_file)
-        self.cache: dict[datetime.date, dict[str, Decimal]] = {
-            **read_data,
-            **(initial_data or {}),
```
vmartinv, line 44: initial_data here is used to supply test data, so it seems we're implementing mocking capabilities twice. I'd say let's keep only one.

vmartinv (changes requested):
Can you elaborate a bit how ERI test cases are specially affected by this?

.env
```diff
@@ -0,0 +1 @@
+CGT_TEST_MODE_STRICT=1
```
KapJI, line 1: I'd rather set it directly in CI command.

tests/exchange_rates_data.csv
```diff
@@ -1,38 +1,24 @@
 month,currency,rate
```
KapJI, line 1: It'd be nice to clearly indicate that this file is generated and how it's supposed to be updated. E.g. you can introduce comments syntax and filter out comments when reading.

cgt_calc/const.py
```diff
@@ -61,7 +62,25 @@
 # General constants
 # =============================================================================
 
-CGT_TEST_MODE = os.environ.get("CGT_TEST_MODE", "0") == "1"
+
+class TestMode(Enum):
```
KapJI, line 66: It's not test mode if one of the options is `PROD`. `RuntimeMode`?

Issue #758, opened 2026-03-14
[Suggestion][Vanguard] support parsing both "Cash" and "Investment" transactions table/csv


newwaylw:
When you export Vanguard Client Transactions (Excel), you get 2 types of transactions per account. They are

"Cash Transactions" which contains fields:

```Date  Details  Amount  Balance```

 and "Investment Transactions" which contains fields:

```Date  InvestmentName  TransactionsDetails  Quantity  Price  Cost```

Current implementation parser vanguard.py assumes the data is from the "Cash Transactions"
The advantage of using "Cash Transactions" is that Fees, interests etc are captured. The problems are:

1. Number of shares bought/sold needs to be parsed from natural language text
2. Some transaction details are missing. For example, You can be selling some investments to cover your fees, however this transaction will ONLY be shown as "Account Fee for the period DD-MM-YYYY to DD-MM-YYYY", without telling you how many shares are disposed of.

Using "Investment Transactions" will give you every bought/sold event and provide the quantity/price without parsing from text.

vmartinv:
Closing since #773 landed

PR #773, opened 2026-03-24
Enhanced vanguard parser

newwaylw:
The Vanguard parser was refactored to support the dual-table CSV format exported from Vanguard report generator:

 - **New table types**: Added CashColumn, InvestmentColumn, and TableType enums to model the two table sections (Cash Transactions and Investment Transactions).
- **Table splitting**: _split_tables() detects and separates the two table sections by matching header columns.
- **Enrichment logic**: Cash transactions are now created first, then enriched via enrich_details() using matched investment rows for missing quantity/price data if any.
- **Investment matching**: _find_investment_match() pairs cash transactions to investment rows by TransactionDetails, with date-aware tie-breaking (earliest after for BUY, latest before for SELL).
- **Delimiter detection**: _detect_delimiter() supports both comma and tab-delimited files.
- **Backward compatibility**: Old single-table (cash-only) CSV files still work via fallback logic. COLUMNS alias preserved for existing tests.

see [[/issue](https://github.com/KapJI/capital-gains-calculator/issues/758)] for more context


cgt_calc/main.py
```diff
@@ -122,7 +122,7 @@ def _approx_equal_price_rounding(
     elif calculation_type is CalculationType.DISPOSAL:
         calculated_amount = quantity_on_record * price_on_record - fees_on_record
         calculated_price = (amount_on_record + fees_on_record) / quantity_on_record
-    in_acceptable_range = abs(calculated_price - price_on_record) < Decimal("0.0001")
+    in_acceptable_range = abs(calculated_price - price_on_record) < Decimal("0.0005")
```
vmartinv: this weakens the check for all brokers but seems reasonable

vmartinv:
@newwaylw  ivendor is busy for some weeks, I can accept the PR once you resolve the conflicts

cgt_calc/main.py
newwaylw: Good call out, actually this change is not needed anymore. It turns out that the discrepancies between Vanguard's two transaction report (Investment  vs Cash) can be more than 0.5, so I kept the cash report number and use it to infer the price so there's no discrpancy. will revert this change in the revision

newwaylw:
@vmartinv here you go

vmartinv:
Thank you!

PR #789, opened 2026-04-20
Vanguard parser: add ETF dealing fee as TRANSFER

jonathangriffiths:
ETF dealing fees currently break parsing of the vanguard output -- this correctly treats them as a fee


vmartinv:
Thanks!

PR #790, opened 2026-04-20
Vanguard parser: exclude index switching events from calcs

jonathangriffiths:
In 2020, a Vanguard ETF that tracked the DAX swapped to tracking a larger (in company number) German index.

If you were affected by this, then the vanguard parser currently breaks. This change attempts to fix this, by ignoring the switch event itself (which is not CGT chargeable), and correcting purchases before the swap to the new name, for correct calculation of purchase price averages.

### DIFF
--- a/cgt_calc/const.py
+++ b/cgt_calc/const.py
@@ -76,6 +76,11 @@ TICKER_RENAMES: Final[dict[str, str]] = {
     "FB": "META",
 }
 
+# Encodes the old ticker in the RENAME transaction's description field;
+# parsed back out by main.apply_ticker_renames. Keep parser + preprocessor
+# in sync.
+RENAME_DESCRIPTION_PREFIX: Final = "renamed from "
+
 
 # =============================================================================
 # Resource files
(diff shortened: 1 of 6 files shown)


vmartinv:
For transparency I maybe is better if we add a new "rename" transaction. Is this too complicated? I imagine we would need to be careful with the Bed&Breakfast rule (i.e. sell->rename->buy back within less than 30 days)

jonathangriffiths:
I think this makes sense, I'll push another commit in a minute. The B&B will be tested there as well

I've tried to avoid any blast zone that extends beyond the vanguard parser as I'm not sure how other platforms handle it

vmartinv:
Sorry I didn't explain it correctly. I don't think we should rewrite past history. I mean that the report should say something like this:
- Bought FOO on day 1
- FOO was renamed to NEW_FOO on day 3
- Sold NEW_FOO on day 5

This is more transparent because it explains exactly what happened with the shares. The rename transaction is just a normal transaction that when processed just moves the shares with its cost to the new name. No rewriting of other transactions.

jonathangriffiths:
Okay, because the events are so historic, I didn't consider that we would be covering the rename date itself -- bit obviously that's a better solution. I'll close this and submit a new one.

PR #803, opened 2026-06-06
Dividend allowance 2025

Ivo-Evans:
Hi, very simple PR to add the 2025 dividend allowance and some missing colons to the output report


KapJI:
Thanks!

PR #808, opened 2026-07-28
test: cover the six HMRC pooling examples

ViieeS:
The suite tests two HS284 examples and two from HS265. It did not cover the six worked pooling examples of CRYPTO22251 to CRYPTO22256, which go through the share identification rules in more detail than HS284 does — they are in the cryptoassets manual only because that is where HMRC chose to work them through; the rules are the ordinary ones and apply unchanged.

  Example 1  a plain section 104 disposal
  Example 2  two disposals and an acquisition on one day, treated as one
  Example 3  the 30 day rule, matched across a tax year end
  Example 4  the same day rule meeting the pool
  Example 5  the 30 day rule meeting the pool
  Example 6  all three rules on one disposal

Example 3 is the one worth having: HMRC's dates put the disposal on 31 March and the acquisitions it matches against in April, so the rule reaches into the following tax year. It is checked from both years, since a run covers one year at a time.

The transactions are HMRC's throughout. Three of the expected gains are the unrounded equivalent of the published figure — HMRC rounds an intermediate cost to whole pounds and this code does not, giving £462.50 against their £462, £79.50 against £79, and £163,636.36 against £163,637. Each is noted in a comment beside the case.

### DIFF
--- a/tests/general/calc_test_data.py
+++ b/tests/general/calc_test_data.py
@@ -168,6 +168,62 @@ def split_transaction(
     )
 
 
+# HMRC's Example 3 spans a tax year end: the disposal of 31 March is matched
+# against acquisitions made in April, which fall in the next year. The same
+# transactions are therefore checked twice, once per year.
+CRYPTO22253_TRANSACTIONS = [
+    transfer_transaction(datetime.date(day=1, month=1, year=2021), 2000),
+    # Opening pool: 2,000 tokens for £1,000.
+    buy_transaction(
+        date=datetime.date(day=1, month=1, year=2021),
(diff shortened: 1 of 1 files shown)


vmartinv:
This is a good addition. Untouched .md is failing because of URL the check, any ideas to fix this @KapJI ? I think it has been around for a while.

ViieeS:
@vmartinv I fixed it in a separate PR: https://github.com/KapJI/capital-gains-calculator/pull/809 )

vmartinv:
CI problem was addressed in #817, this is safe to land. Thank you!

PR #812, opened 2026-08-03
fix: parse ordinary dividends and FX adjustments from Interactive Brokers

ViieeS:
A real IBKR Transaction History statement failed on unsupported `Adjustment` and `Dividend` rows.
The parser now:

- Supports ordinary dividends.
- Maps FX translation adjustments to `ActionType.ADJUSTMENT` as non-taxable cash movements.
- Treats `"-"` as empty in text fields, matching numeric fields.

Fixtures and tests now cover dividends, withholding tax, and FX adjustments. A real statement parses successfully, with dividend and tax totals matching IBKR’s dividend report and 1042-S.


vmartinv:
Thank you!

PR #814, opened 2026-08-04
Warnings are not scoped to the tax year being computed

ViieeS:
## Problem

`--year 2025` emits 15 warnings, fewer than half about 2025. The calculator must walk the whole history, but it warns as if every year were in scope. Users learn to scroll past the block, so warnings that matter get missed.

## Change

Gate the treaty-mismatch warning on the computed tax year. Dividends outside it are skipped entirely — they were never reported either, since `treaty` and the `Dividend` entry are already built under a `date_in_tax_year` check.

Two tidy-ups in the same area:

- Bed and breakfasting is a record of how a disposal was matched, not a problem → `WARNING` to `INFO`.
- `No Schwab Award file provided` now fires on the first row that needs a price, not before any row is parsed. Today a cash-only Schwab account triggers it with nothing to price. It still fires at most once, since the following `AwardPrices.get` raises.

## Verification

Real 2025/26 run across Schwab equity award JSON, Schwab CSV, and IB CSV: stdout byte-identical, stderr 16 warnings → 0.

Six tests asserted warnings that are now silent. None of those fixtures has a `Stock Plan Activity` row, so `No Schwab Award file provided` was spurious in all six.

## Open questions

- Warnings about other years could go behind an opt-in flag; left out to keep this small.
- When the award file is genuinely needed and absent, the warning is followed by an unhandled `KeyError` from `AwardPrices.get`. A `ParsingError` naming `--schwab-award-file` would read better — here or in a separate PR?


cgt_calc/parsers/schwab.py
```diff
@@ -256,6 +256,10 @@ def create(
             symbol = transaction.symbol
             if symbol is None:
                 raise SymbolMissingError(transaction)
+            if not awards_prices:
+                # Only rows priced from the awards file need it, and this is the
+                # first of them: an account without equity awards never gets here.
+                LOGGER.warning("No Schwab Award file provided")
```
vmartinv: since the warning is now within a specific transaction in context can you add information about it in the log? (i.e. date, symbol, etc)

vmartinv (changes requested):

vmartinv:
The ParsingError naming --schwab-award-file sounds good, you can add it in a separate PR after we merge this one.
Or also you can add it here if it's more convenient.

ViieeS:
Done — the warning now names the symbol, quantity and date of the stock activity it could not price.

vmartinv:
thanks!

PR #815, opened 2026-08-04
Tax treaty selected by transaction currency instead of the income's source country

ViieeS:
## Problem

Two US-domiciled securities with identical 15% US withholding are treated differently in one run: `NVDA` gets treaty relief, `VT` does not. The reported taxable dividend total is wrong as a result.

Currency is a property of the broker, not of the income. Schwab reports US dividends in USD; Interactive Brokers converts them to the account's base currency, GBP for a UK account. Two things go wrong:

- `DIVIDEND_DOUBLE_TAXATION_RULES` is keyed by currency and `main.py` looks up `foreign_amount.currency`.
- For a GBP-denominated foreign dividend the lookup is never reached at all, because the block sits under `elif foreign_amount.currency != UK_CURRENCY`. That guard is also why `Taxation treaty for %s country is missing` never fires for such a dividend.

The guard looks redundant: the enclosing `if tax.amount < 0` already requires withholding at source, which a genuine UK dividend does not have.

## Change

The source country now comes from the ISIN, which is what the codebase already uses to identify a security — `BrokerTransaction` has the field and `IsinConverter.add_from_transaction` collects the mapping from every transaction. No broker parser populated it, so the IBKR parser starts reading it from the description it is prefixed with (`VT(US9220427424) Cash Dividend ...`), validated with `is_isin`.

- `DIVIDEND_DOUBLE_TAXATION_RULES` is re-keyed by country code. `TaxTreaty` already carried the country, so the currency key was the odd one out. It is a module constant with one call site and no CLI surface, so nothing user-facing changes shape.
- Brokers that supply no ISIN fall back to guessing the country from the currency, keeping today's behaviour for them.
- The `!= UK_CURRENCY` guard is removed.

I avoided adding a ticker → country mapping: it would duplicate what ISIN already models, and a hardcoded one does not scale, since every user holds different tickers.

## Verification

`pytest`, `ruff`, `mypy` clean. Seven new tests: five covering each way the country is resolved — three of which fail without this change, two pinning the currency fallback — and two on the parser.

On a real 2025/26 run across Schwab equity award JSON, Schwab CSV and IB CSV, `VT` now gets relief alongside `NVDA`, and the reported taxable dividend total moves from £29.99 to £0.00: `max(0, 580.54 − 500.00 − (50.55 + 36.53))`.

Two behaviour changes worth calling out:

- The fixture's withholding on a US ISIN was 20%, which the calculator now rightly reports as not matching the 15% treaty rate. I changed the synthetic figure to 15% so the fixture exercises relief being granted; no logic depends on it.
- `Source country of the %s dividend is unknown` is a new warning for anyone whose broker gives no ISIN and pays in GBP with tax withheld. Those cases were silent before.

## Open questions

- Is the currency fallback worth keeping at all, or would you rather every parser be taught to supply an ISIN first? I kept it so no existing setup regresses.
- Only `US` and `PL` exist in the table today. If you would like more treaties added while the keys are being rewritten anyway, say which.

### DIFF
--- a/cgt_calc/const.py
+++ b/cgt_calc/const.py
@@ -50,12 +50,25 @@ DIVIDEND_ALLOWANCES: Final[dict[int, int]] = {
 # Double taxation
 # =============================================================================
 
-# Country and treaty rates per country
+# Country and treaty rates per country of the income's source, keyed by the
+# ISO 3166-1 alpha-2 code that an ISIN is prefixed with.
 # https://www.gov.uk/hmrc-internal-manuals/double-taxation-relief
 DIVIDEND_DOUBLE_TAXATION_RULES: Final[dict[str, TaxTreaty]] = {
-    "USD": TaxTreaty("USA", Decimal("0.15"), Decimal("0.15")),
-    "PLN": TaxTreaty("Poland", Decimal("0.19"), Decimal("0.1")),
+    "US": TaxTreaty("USA", Decimal("0.15"), Decimal("0.15")),
+    "PL": TaxTreaty("Poland", Decimal("0.19"), Decimal("0.1")),
 }
+
+# Fallback for transactions with no ISIN: guess the source country from the
+# currency the dividend was paid in. This is only a guess — a broker reporting
+# in the account's base currency breaks it — so it is used only as a last
+# resort, and it keeps the behaviour brokers relied on before ISINs were used.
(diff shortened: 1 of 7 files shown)


vmartinv:
these changes are correct, could you please resolve the conflicts?

ViieeS:
Rebased on main, conflicts resolved.

vmartinv:
Thank you!

PR #817, opened 2026-08-06
Upload latex artifacts for debugging & fix latex race condition issues while testing

vmartinv:
There are transient issues in the CI, I believe this is because all our e2e integration tests attempt to write into the same out/ folder. I have now:
- Added isolated output dirs for each of these tests
- Added a CI step to upload these outputs so is easier to debug CI e2e issues in the future

If you go to the "View details" in the added step, the logs provide the download URL. The outputs are only 667KB.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -77,4 +77,15 @@ jobs:
 
       - name: Run pytest inside container
         run: |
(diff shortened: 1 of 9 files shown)


vmartinv:
Adding this to fix CI in main

KapJI:
Thanks for looking into this!

PR #818, opened 2026-08-07
Missing Schwab award file ends in a traceback instead of a message

ViieeS:
Follow-up to #814, as suggested there.

## Problem

A Stock Plan Activity row with no price of its own is priced from the Equity Awards file. When that lookup fails, the `KeyError` from `AwardPrices.get` was never caught, so the run ended in a traceback:

```
CRITICAL: Unexpected error!
ERROR: Details:
Traceback (most recent call last):
  ...
KeyError: 'Award price is not found for symbol BAR for date 2023-08-18'
```

Nothing in that says the user simply forgot `--schwab-award-file`.

## Change

`ParsingError` instead, naming the vest and what to do about it:

```
ERROR: While parsing .../transactions.csv: Cannot price a vest: the BAR stock
activity of 200 on 2023-08-18 has no price of its own, and no Schwab Award file
was provided. Pass the Equity Awards transaction history with --schwab-award-file.
```

A file that was provided but does not cover this award reads `the Schwab Award file has no award price for it` instead, since that is a different mistake.

The `No Schwab Award file provided` warning is removed. Since #814 it fired only on a row that needs the file, and the very next statement is the lookup that now fails — so it only ever appeared immediately before this error, and everything it reported is in the message.

## Verification

`pytest`, `ruff`, `mypy` clean. Two tests, one per reason the lookup can fail; both assert the vest is named.

### DIFF
--- a/cgt_calc/parsers/schwab.py
+++ b/cgt_calc/parsers/schwab.py
@@ -256,21 +256,30 @@ class SchwabTransaction(BrokerTransaction):
             symbol = transaction.symbol
             if symbol is None:
                 raise SymbolMissingError(transaction)
-            if not awards_prices:
-                # Only rows priced from the awards file need it, and this is the
-                # first of them: an account without equity awards never gets here.
-                LOGGER.warning(
-                    "No Schwab Award file provided, needed to price the %s "
-                    "stock activity of %s on %s",
-                    symbol,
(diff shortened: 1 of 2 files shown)


vmartinv:
Looks great! Thank you

PR #819, opened 2026-08-08
Ease out updating of test exchange_rates when ERI or new test cases are added - Take 2

ivendor:
When introducing new ERI values or new broker test case are added we have to now manually append entries into exchange rates. To workaround this problem, I've forked the test mode in 2 test modes:

TEST MODE when the pytest is run outside of pre-commit hook which auto append entries to the test mocks supporting parallel execution of tests via file locks
TEST STRICT MODE when the pytest is run inside the pre-commit hook which beahves like TEST MODE before this change.
Finally regenerated the CSV file using TEST MODE

Updated with comments from:
https://github.com/KapJI/capital-gains-calculator/pull/760


ivendor:
cc @vmartinv

cgt_calc/currency_converter.py
```diff
@@ -210,6 +243,18 @@ def currency_to_gbp_rate(self, currency: str, date: datetime.date) -> Decimal:
             self._query_hmrc_api(date)
         if currency not in self.cache[date]:
             raise ExchangeRateMissingError(currency, date)
+
+        if CGT_MODE == RuntimeMode.TEST and self.exchange_rates_file is not None:
+            if date not in self.test_cache:
+                self.test_cache[date] = {}
+            if currency not in self.test_cache:
```
vmartinv: compares a currency string against a dict keyed by date, so it's basically always true.

cgt_calc/currency_converter.py
```diff
@@ -39,10 +40,19 @@ def __init__(
         """Load data from exchange_rates_file and optionally from initial_data."""
         self.exchange_rates_file = exchange_rates_file
         read_data = self._read_exchange_rates_file(exchange_rates_file)
-        self.cache: dict[datetime.date, dict[str, Decimal]] = {
-            **read_data,
-            **(initial_data or {}),
-        }
+        self.cache: dict[datetime.date, dict[str, Decimal]] = {}
```
vmartinv: self.cache starts empty in TEST mode, so every date is treated as a miss and hits the live HMRC API every time, even for dates already in the fixture — test_cache is never consulted before fetching.

cgt_calc/currency_converter.py
```diff
@@ -39,10 +40,19 @@ def __init__(
...
+            self.test_cache = {
+                **read_data,
+                **(initial_data or {}),
+            }
+        else:
+            self.cache = {
+                **read_data,
+                **(initial_data or {}),
+            }
+
```
vmartinv, line 48: I don't like having two caches here. I think we are mixing prod and test logic too much. Can we create a `class TestRecorderCurrencyConverter(CurrencyConverter)` with the additional logic. Switching CGT_MODE should select the appropiate class.

vmartinv (changes requested):

cgt_calc/currency_converter.py
```diff
@@ -52,98 +55,119 @@ def __init__(
...
-            return cache
+            return CurrencyConverter._read_exchange_rates_data(exchange_rates_file, fin)
 
     @staticmethod
     def _write_exchange_rates_file(
         exchange_rates_file: Path | None, data: dict[datetime.date, dict[str, Decimal]]
     ) -> None:
-        if exchange_rates_file is None or CGT_TEST_MODE:
+        if not exchange_rates_file or CGT_MODE != RuntimeMode.PROD:
             return
```
vmartinv: the test classes need to override this function, so this CGT_MODE check can be removed.

vmartinv (changes requested):
looks much better now, one minor detail

vmartinv:
Great thank you! I will add the 2025 data shortly

PR #820, opened 2026-08-09
Add HL parser

ivendor:
Adding new parser for HL: https://www.hl.co.uk/

HL exports tax reports in CSV format (for operation list) + PDF format (for full data on buy/sell operation).
The parser does both CSV and PDF parsing.
I do not have dividend example hence I didn't add the DIVIDEND transaction type.

Note that I don't have a SELL example so the SELL contract note is based on references found in third party forums: https://forum.portfolio-performance.info/t/pdf-import-from-hargreaves-landsdown/21726

The test PDF used are not real ones but they're generated with a script just to make it compatible with the parsing to avoid sharing personal information

Account management fees (annual charges) are not allowed costs so including them as TRANSFER operations to preserve correctness of balance calculations:
https://www.gov.uk/hmrc-internal-manuals/capital-gains-manual/cg15250


vmartinv:
do we need both CSV and PDF parser for this?

cgt_calc/parsers/hl.py
```diff
@@ -0,0 +1,203 @@
...
+        amount_str = row.get("Value (£)", "0").replace(",", "")
+        amount = (
+            Decimal(amount_str) if amount_str and amount_str != "n/a" else Decimal(0)
+        )
+
+        # Live PDF lookup only when needed for buy/sell actions[cite: 1]
+        if action_type in [ActionType.BUY, ActionType.SELL]:
+            matching_pdfs = list(file_path.parent.glob(f"{reference}_*.pdf"))
+            if matching_pdfs:
+                pdf = cls._parse_pdf(matching_pdfs[0])
```
vmartinv, line 174: can you add a cache so we dont parse the pdf on every transaction? Also is there any limit on the dates? if there is a limit you need to support multiple pdfs (parse and combine them)

cgt_calc/parsers/hl.py
```diff
@@ -0,0 +1,203 @@
...
+    date: date | None
+    action: str | None
+    ticker: str | None
+    quantity: Decimal
+    price: Decimal
+    consideration: Decimal
+    fees: Decimal
+
+
+class HLParser(StandardCSVParser, BaseDirParser):
```
vmartinv: Let's avoid using HL, many are not familiar with this broker.
```suggestion

class HargreavesLansdownParser(StandardCSVParser, BaseDirParser):
```

tests/hl/test_hl.py
```diff
@@ -0,0 +1,46 @@
...
+        / "data"
+        / "test_run_with_hl_files_no_balance_check_output.txt"
+    )
+    expected = expected_file.read_text()
+    cmd_str = " ".join([param or "''" for param in cmd])
+    assert result.stdout == expected, (
+        "Run with example files generated unexpected outputs, "
+        "if you added new features update the test with:\n"
+        f"{cmd_str} > {expected_file}"
+    )
```
vmartinv, line 122: uhm this test is just testing that it failed, run the suggested command

cgt_calc/parsers/hl.py
```diff
@@ -0,0 +1,203 @@
...
+
+        # Live PDF lookup only when needed for buy/sell actions[cite: 1]
+        if action_type in [ActionType.BUY, ActionType.SELL]:
+            matching_pdfs = list(file_path.parent.glob(f"{reference}_*.pdf"))
+            if matching_pdfs:
+                pdf = cls._parse_pdf(matching_pdfs[0])
+                symbol = pdf.ticker
+                quantity = pdf.quantity
+                price = pdf.price / Decimal(100)
+                fees = pdf.fees
```
vmartinv, line 179: if there is no matching pdf the transaction will be incomplete:
1. maybe raise an error for this
2. add a test case with missing PDF to make sure we are failing

vmartinv (changes requested):

cgt_calc/parsers/hl.py
vmartinv: oh nevermind, I get it now , there is a PDF for each transaction? that is really cumbersome. then the current code is fine.

ivendor: Yes the broker provides you a CSV report for all tax relevant operations and then single PDFs for each buy and sell which contains the detail data

ivendor:
> do we need both CSV and PDF parser for this?

yes unfortunately HL needs both the CSV for the general transactions and PDF for the buy and sell operations

tests/hl/test_hl.py
ivendor: what do you mean? The test has a successfull buy and sell transaction which result in the outcome, which is correctly verified (if I empty the outcome txt it fails). I'll though change it slightly to generate capital gains (currently it doesn't) and then add a test for what you suggested above

ivendor:
removed hardcoded generated PDF and added inline test PDF generation

tests/hl/test_hl.py
vmartinv: Maybe the message is not clear? It says the tests generated unexpected outputs and failed. It suggests you do run a command to fix it (the command to generate the expected output). If this test is running example transactions it shouldn’t fail.

ivendor: The message is for when the assertion (result.stdout != expected) hence the test fails, it just provide you instructions on how to change the test expected file to make it succeed. It doesn't show up when test succeed

vmartinv: Could you please run the command. There shouldn’t be any unexpected tests. You are just missing the expected output file.

ivendor: I don't think I'm missing anything, when I started writing the test I did it with empty output expected file, if I do this now (i.e. I manually empty the txt expected file which is in this PR) I get:

Screenshot 2026-08-11 at 08 58 34

If I then run the generated command and I re-run the tests

Screenshot 2026-08-11 at 09 00 12

vmartinv: could you compare with the other brokers? this is not a successful test, you are generating errors

ivendor: Other broker with similar assertions (this is where I took inspiration from): 

https://github.com/KapJI/capital-gains-calculator/blob/main/tests/vanguard/test_vanguard.py#L45

note that I'm not generating any error the test passes cleanly or CI won't allow be saying ready to merge.

Maybe I'm missing something but I don't get how you come to the conclusion that this is not a successful test. If the output matches perfectly and there are buy and sell transactions and CGT calculations.

To recap the test is asserting in order:
- Return Code is 0 (success) `if result.returncode:`
- that result stderr is empty `assert result.stderr == ""`
- that output matches the expected output txt `assert result.stdout == expected`

vmartinv: sorry I just saw it now... mmm maybe we can put this logic in a utility function.

vmartinv:
Thank you, and sorry for the confusion

PR #847, opened 2026-08-11
Fix Schwab Cancel Buy matching direction for newest-first exports

vmartinv:
## Summary

`_filter_cancelled_buy_transactions` (in `cgt_calc/parsers/schwab.py`) matches
each "Cancel Buy" row to the original Buy transaction it cancels by searching
the raw transaction list (before `transactions.reverse()` is applied) around
the Cancel Buy's index within a date window.

It searched **backward** (`range(cancel_idx - 1, -1, -1)`), which is only
correct if the list is ordered oldest-first. Real Schwab CSV exports are
ordered **newest-first** — see `tests/schwab/data/schwab_transactions.csv`,
where 2021 transactions appear at the top of the file and 2020/2016
transactions appear at the bottom.

Since a Cancel Buy is chronologically *after* the Buy it cancels, in a
newest-first list the Cancel Buy row has a **lower** index than its matching
Buy row. Searching backward from the Cancel Buy's index therefore only looks
at transactions that are *more recent* than the Cancel Buy — it can never
reach the Buy it's supposed to match, so on real exports the Cancel Buy is
never matched, and both the cancelled Buy and the Cancel Buy row are left in
the transaction list. This corrupts capital gains calculations by including a
purchase that was actually cancelled.

## Fix

Search **forward** instead (`range(cancel_idx + 1, len(transactions))`), i.e.
toward higher indices / older transactions, which is the correct direction
for a newest-first list. Updated the docstring/comments to explain the
ordering assumption explicitly, so this doesn't regress again.

## Tests

- Reordered the CSV fixtures in the existing `test_schwab_cancel_buy.py`
  tests to reflect the realistic newest-first ordering (Cancel Buy row above
  its Buy row), matching how Schwab actually exports data.
- Added a new regression test,
  `test_cancel_buy_after_original_buy_in_newest_first_order`, with an extra
  unrelated older transaction in the mix, which fails under the old backward
  search and passes with the fix.

## Test plan

- [x] `CGT_TEST_MODE_STRICT=1 uv run pytest` — 244 passed
- [x] `uv run mypy .` — no new errors (one pre-existing, unrelated error in
      `scripts/import_eri_reports.py` present on `main` as well)
- [x] `uv run ruff format --check .` / `uv run ruff check .` — clean
- [x] `uv run pylint cgt_calc tests` — 10.00/10
- [x] pre-commit hooks (ruff, mypy, pylint, pytest, etc.) all pass

### DIFF
--- a/cgt_calc/parsers/schwab.py
+++ b/cgt_calc/parsers/schwab.py
@@ -495,8 +495,15 @@ def _filter_cancelled_buy_transactions(
     This is a Schwab-specific quirk - other brokers may not report cancellations
     at all or may handle them differently.
 
+    Note: this runs on the raw, newest-first ordered list of transactions as
+    read from the Schwab export (before ``transactions.reverse()`` is applied
+    by the caller). A Cancel Buy row is chronologically *after* the Buy it
+    cancels, so in a newest-first list the Cancel Buy has a *lower* index than
+    its Buy. We therefore search forward (increasing index = older dates) from
+    the Cancel Buy's index.
+
     Args:
-        transactions: List of parsed Schwab transactions
+        transactions: List of parsed Schwab transactions, newest-first
 
     Returns:
         Filtered list with Cancel Buy pairs removed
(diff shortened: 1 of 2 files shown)


vmartinv:
@amaksoft please see fix to bug introduced from #686 , can you confirm this since I don't have any Cancel Buy transactions?

amaksoft:
Hmm.. makes sense to me @vmartinv

PR #849, opened 2026-08-11
Fix unrealized gains conversion hardcoding USD as ticker currency

vmartinv:
## Summary

`CurrentPriceFetcher.get_current_market_price()` (used by `--calc-unrealized-gains`) fetched the price from `yf.Ticker(symbol).info` but always converted it to GBP as if it were quoted in USD:

```python
market_price_usd = Decimal(format(market_price, ".15g"))
return self.converter.to_gbp(
    market_price_usd, "USD", datetime.datetime.now().date()
)
```

`yfinance`'s `ticker.info` actually exposes a `"currency"` field describing what currency the price is quoted in (e.g. `"USD"`, `"EUR"`, or `"GBp"` for LSE-listed stocks quoted in pence). Ignoring it means:

- A **EUR-quoted** ticker gets converted using the USD/GBP rate instead of the EUR/GBP rate — wrong conversion factor.
- A **GBp-quoted** (LSE) ticker, e.g. `regularMarketPrice = 100` meaning 100 pence = £1, gets divided by the USD/GBP rate instead of simply divided by 100 — off by roughly 100x. Note `"GBp"` is a yfinance-specific quirk, not a real ISO 4217 code, so `CurrencyConverter` (which only knows real ISO currency codes/HMRC rates) can't and shouldn't be asked to look up a rate for it.

This silently corrupts the `--calc-unrealized-gains` report for any non-USD symbol, which is plausible for HL/Freetrade users holding LSE tickers.

## Fix

- Added a `_convert_to_gbp` helper on `CurrentPriceFetcher` that:
  - Special-cases `currency == "GBp"` by dividing by 100 directly (no FX lookup, since it's already GBP-equivalent pence).
  - Otherwise passes the ticker's real `currency` field through to `CurrencyConverter.to_gbp(...)`.
  - Falls back to `"USD"` only when the `currency` field is genuinely absent, preserving prior behavior for USD tickers.
- `get_current_market_price()` now reads `ticker.get("currency")` from `.info` and uses the helper instead of hardcoding `"USD"`.
- `get_closing_price()` had the identical hardcoded-`"USD"` bug for historical closing prices. Fixed it too: `yf.Ticker(symbol).history(...)` doesn't carry a currency column, but the same `Ticker` object's `.info["currency"]` does, so the method now fetches `.info` once alongside `.history(...)` and reuses the same helper.

## Test plan

- Added tests in `tests/general/test_current_price_fetcher.py` (following the existing `FakeTicker`/`monkeypatch` pattern):
  - `test_gbp_pence_quoted_ticker_is_divided_by_100` — GBp price of 100 converts to £1.00, no FX lookup.
  - `test_eur_quoted_ticker_uses_eur_rate_not_usd` — EUR-quoted ticker converts using the EUR rate, not USD.
  - `test_defaults_to_usd_when_currency_field_absent` — existing USD-default fallback behavior still works when `currency` is missing.
- [x] `CGT_TEST_MODE_STRICT=1 uv run pytest` — all pass (one pre-existing, unrelated failure in `tests/hl/test_hl.py::test_hl_parser` reproduces identically on unmodified `main`, likely an environment/PDF-tooling issue).
- [x] `uv run mypy cgt_calc tests`
- [x] `uv run ruff format --check` / `uv run ruff check`
- [x] `uv run pylint cgt_calc tests` (10.00/10)
- [x] pre-commit hooks pass on the commit

### DIFF
--- a/cgt_calc/current_price_fetcher.py
+++ b/cgt_calc/current_price_fetcher.py
@@ -27,6 +27,20 @@ class CurrentPriceFetcher:
         self.historical_prices_data = historical_prices_data or {}
         self.converter = converter
 
+    def _convert_to_gbp(self, price: Decimal, currency: str | None) -> Decimal:
+        """Convert a price quoted in the given yfinance currency to GBP.
+
+        yfinance uses the non-ISO code "GBp" (lowercase p) for LSE-listed
+        instruments quoted in pence rather than pounds, so that case is
+        converted directly instead of going through CurrencyConverter, which
+        only knows real ISO currency codes and has no exchange rate for it.
+        """
+        if currency == "GBp":
+            return price / Decimal(100)
+        return self.converter.to_gbp(
+            price, currency or "USD", datetime.datetime.now().date()
+        )
+
     def get_current_market_price(self, symbol: str) -> Decimal | None:
         """Given a symbol gets the current market price."""
         if self.current_prices_data is not None and symbol in self.current_prices_data:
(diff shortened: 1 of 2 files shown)


vmartinv:
Verified live against the real yfinance API (not just the mocked tests) that the `currency` field and values this fix relies on are accurate:

```
VOD.L -> {'currency': 'GBp', 'currentPrice': 118.3, 'regularMarketPrice': 118.3, 'navPrice': None}
AAPL  -> {'currency': 'USD', 'currentPrice': 304.91, 'regularMarketPrice': 304.91, 'navPrice': None}
MC.PA -> {'currency': 'EUR', 'currentPrice': 478.8, 'regularMarketPrice': 478.8, 'navPrice': None}
```

Confirms `"GBp"` (pence, not a real ISO code) is indeed what yfinance returns for LSE-listed tickers like `VOD.L`, and that `AAPL`/`MC.PA` report standard `"USD"`/`"EUR"` — matching the assumptions this fix's `_convert_to_gbp` helper and tests are built on.

PR #853, opened 2026-08-12
Add support for processing file descriptors and standard input

### LINKED ISSUE #788: Support process substitution for input files?
I tried to run the calculator on the output of another tool to produce the transactions, and it rejected the generated filename:

```
; cgt-calc --raw-file <(my-command ...)
usage: cgt-calc [--year YYYY] [--freetrade-file PATH] [--raw-file PATH] [--schwab-award-file PATH] [--schwab-file PATH] [--schwab-equity-award-json PATH]
                [--sharesight-dir DIR] [--trading212-dir DIR] [--mssb-dir DIR] [--vanguard-file PATH] [--interactive-brokers-file PATH] [--eri-raw-file PATH]
                [--initial-prices-file PATH] [--exchange-rates-file PATH] [--isin-translation-file PATH] [--spin-offs-file PATH] [--no-balance-check]
                [--unrealized-gains] [--interest-fund-tickers TICKER[,TICKER...]] [-o PATH | --no-report] [-h] [--version] [-v]
cgt-calc: error: argument --raw-file: expected file path, got: '/proc/self/fd/11'
```

which looks like this check: https://github.com/KapJI/capital-gains-calculator/blob/fb964592b9b4e4c8dbbe2253913101b53f15e869/cgt_calc/args_validators.py#L93-L94

It would be very convenient if this could be relaxed. Of course the program would have to read the data only once, buffering internally if necessary.

Even more convenient would be to support `cgt-calc --raw-file -` for standard input :)

dwoffinden:
E.g. `cgt-calc --raw-file <(my-command ...)` or `my-command ... | cgt-calc --raw-file -`

Fixes https://github.com/KapJI/capital-gains-calculator/issues/788

### DIFF
--- a/cgt_calc/args_validators.py
+++ b/cgt_calc/args_validators.py
@@ -10,6 +10,8 @@ from typing import TYPE_CHECKING
 
 from .const import INTERNAL_START_DATE
 
+STDIN_PATH = Path("-")
+
 if TYPE_CHECKING:
     from collections.abc import Sequence
     from typing import Any
@@ -73,9 +75,11 @@ def optional_file_type(value: str) -> Path | None:
     """Convert non-empty value to Path and ensure file semantics."""
     if value.strip() == "":
         return None
+    if value == "-":
+        return STDIN_PATH
     path = Path(value)
     if path.exists():
-        if not path.is_file():
+        if not (path.is_file() or path.is_fifo()):
             raise argparse.ArgumentTypeError(
                 f"expected file path, got directory: '{value}'"
             )
@@ -90,7 +94,7 @@ def _existing_path_type(value: str, *, require_dir: bool) -> Path:
         raise argparse.ArgumentTypeError(f"path does not exist: '{value}'")
     if require_dir and not path.is_dir():
         raise argparse.ArgumentTypeError(f"expected directory path, got: '{value}'")
-    if not require_dir and not path.is_file():
+    if not require_dir and not (path.is_file() or path.is_fifo()):
         raise argparse.ArgumentTypeError(f"expected file path, got: '{value}'")
     if require_dir:
         _ensure_readable_directory(path, value)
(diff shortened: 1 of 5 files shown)


cgt_calc/args_validators.py
```diff
@@ -86,21 +88,23 @@ def optional_file_type(value: str) -> Path | None:
 def _existing_path_type(value: str, *, require_dir: bool) -> Path:
     """Ensure provided path exists and matches expected type."""
     path = Path(value).expanduser()
-    if not path.exists():
-        raise argparse.ArgumentTypeError(f"path does not exist: '{value}'")
-    if require_dir and not path.is_dir():
-        raise argparse.ArgumentTypeError(f"expected directory path, got: '{value}'")
-    if not require_dir and not path.is_file():
```
vmartinv, line 93: we shouldn't remove the is_file() check, we need to check whether is file OR a FD path.

tests/raw/test_raw.py
```diff
@@ -81,6 +81,41 @@ def test_run_with_raw_files() -> None:
...
+    expected = expected_file.read_text()
+
+    cmd_str = " ".join([param or "''" for param in cmd])
+    assert result.stdout == expected, (
+        "Run with stdin generated unexpected outputs, "
+        "if you added new features update the test with:\n"
+        f"cat {csv_file} | {cmd_str} > {expected_file}"
+    )
+
+
```
vmartinv, line 118: can you add a test where a non-existing file is provided and the correct error gets raised?

vmartinv (changes requested):

cgt_calc/args_validators.py
dwoffinden: isn't that covered by `_ensure_readable_file()`?

tests/raw/test_raw.py
dwoffinden: done

cgt_calc/args_validators.py
vmartinv: Partially. ensure_readable_file makes sure the file is readable (i.e. file exists but the tool lacks read permission to it). Having it say the file doesn't exist is a more clear message to me.

dwoffinden: done

vmartinv:
Thank you!

PR #856, opened 2026-08-16
Withholding written before its dividend fails the balance check

ViieeS:
## Problem

IBKR lists tax withheld at source before the dividend it came from. A statement that does not reach back to the opening of the account starts at zero, so the withholding is the first row processed:

```
ERROR: Reached a negative balance(-8.5861881) for broker Interactive Brokers (GBP)
```

The dividend covering it is the next row, same date. #845 does not reach this: both rows sort to the same key.

## Change

`DIVIDEND_TAX` joins `BUY` at the end of the day in `_by_date_and_action`, for the same reason buys are there.

## Verification

`pytest`, `ruff`, `mypy` clean. New regression test fails without the change.

A real 2025/26 run over Schwab equity award JSON, Schwab CSV and IB CSV now completes without `--no-balance-check`, which it could not before. Output with and without the flag is byte-identical.

### DIFF
--- a/cgt_calc/parsers/interactive_brokers.py
+++ b/cgt_calc/parsers/interactive_brokers.py
@@ -224,13 +224,18 @@ class InteractiveBrokersParser(StandardCSVParser):
 
         # If there's a deposit in the same second as a buy
         # (happens with the referral award at least)
-        # we want to put the buy last to avoid negative balance errors
-        return (transaction.date, transaction.action == ActionType.BUY)
+        # we want to put the buy last to avoid negative balance errors.
+        # Tax withheld at source goes last for the same reason: IBKR lists it
+        # before the dividend it was taken from.
+        return (
+            transaction.date,
+            transaction.action in (ActionType.BUY, ActionType.DIVIDEND_TAX),
+        )
 
     @classmethod
     def post_process_transactions(
         cls, transactions: list[BrokerTransaction]
     ) -> list[BrokerTransaction]:
-        """Sort transactions by date, buys last."""
+        """Sort transactions by date, buys and withheld tax last."""
         transactions.sort(key=cls._by_date_and_action)
         return transactions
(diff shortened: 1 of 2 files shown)


cgt_calc/main.py
```diff
@@ -642,7 +644,13 @@ def convert_to_hmrc_transactions(
...
                     + "\n"
                 )
-                msg += "Tip: If your input file is missing deposits/withdrawals use --no-balance-check."
+                msg += (
+                    "Tip: a statement that does not go back to the opening of the "
+                    "account starts the balance at zero, so the first withdrawal "
+                    "reaches this point with the deposits that funded it missing. "
+                    "That is what --no-balance-check is for. Where the whole history "
+                    "is present, this is a genuine gap in the input."
+                )
```
vmartinv: please revert, the original message is more clear.

vmartinv (changes requested):
Please update IBKR parser's post_process_transactions to add the correct order instead. It should be a one/two line fix there. Update the ordering function: instead of `== BUY` it should be `in (BUY, DIVIDEND)`

cgt_calc/main.py
ViieeS: Reverted.

ViieeS:
Moved into `post_process_transactions` as suggested — with `DIVIDEND_TAX` rather than `DIVIDEND`: the key sorts `True` last, so listing `DIVIDEND` there puts the dividend after its withholding and the balance still goes negative. Checked both on the real statement.

Two consequences worth noting: the withholding now goes to the end of the day rather than immediately behind its dividend, and this covers IBKR only — Schwab can write the pair the same way round.

vmartinv:
Thanks! Only the brokers that send the order wrong need fixing.

KapJI:
@ViieeS Offtopic qustion: Are you interested in merging your Nvidia equity-award support as well?

ViieeS:
@KapJI yes) Just want to prepare it and merge minor things first. Will prepare a PR today

ViieeS:
@KapJI ready https://github.com/KapJI/capital-gains-calculator/pull/876

PR #857, opened 2026-08-17
Handle missing TransactionDetails in Vanguard

jonathangriffiths:
I see that old transactions from my recent Vanguard exports are missing the content of a column `TransactionDetails` that explains the transactions in words (you can see what it should look like in [this row](https://github.com/KapJI/capital-gains-calculator/blob/c2406be17e40e822a362fd974cf150afd01f34d8/tests/vanguard/data/cash_investment_report.csv#L32) in the test fixture).

Currently the parser fails completely for these and crashes the whole run - this PR infers the content that would otherwise have been carried in that bit of text from the other columns.

I suspect this is something that has changed on Vanguard's side as I think this hasn't been a problem to date.


cgt_calc/parsers/vanguard.py
```diff
@@ -268,18 +277,22 @@ def _make_transaction_from_investment(
             is_reversal=False,
             description=f"{RENAME_DESCRIPTION_PREFIX}{old_ticker}",
         )
```
vmartinv: this needs to be moved to within the else branch below

vmartinv (changes requested):

cgt_calc/parsers/vanguard.py
jonathangriffiths: Done

vmartinv:
Thanks!

PR #871, opened 2026-08-17
Accept the new Freetrade CSV export header

### LINKED ISSUE #800: Freetrade format change
Looks like freetrade has changed its CSV format:

```csv
Title,Type,Timestamp,Account Currency,Total Amount in Account Currency,Buy / Sell,Ticker,ISIN,Price per Share in Account Currency,Stamp Duty,Quantity,Venue,Order ID,Order Type,Instrument Currency,Total Amount in Instrument Currency,Price per Share,FX Rate,Base FX Rate,FX Fee (BPS),FX Fee Amount,Dividend Ex Date,Dividend Pay Date,Dividend Eligible Quantity,Dividend Amount Per Share,Dividend Gross Distribution Amount,Dividend Net Distribution Amount,Dividend Withheld Tax Percentage,Dividend Withheld Tax Amount,Stock Split Ex Date,Stock Split Pay Date,Stock Split New ISIN,Stock Split Rate of Share Outturn From,Stock Split Rate of Share Outturn To,Stock Split Maintain Holding of Initial ISIN,Stock Split New Share Quantity,Stock Split Rate of Cash Outturn Amount,Stock Split Rate of Cash Outturn Currency,Stock Split Cash Outturn Received Amount,Stock Split Has Fractional Payout,Stock Split Rate of Fractional Payout Amount,Stock Split Rate of Fractional Payout Currency,Stock Split Fractional Payout Cash Received Amount,Stock Split Fractional Payout Cash Received Currency
```

hdimer:
Fixes #800.

Freetrade renamed two columns and added a block of `Stock Split *` ones, so any export downloaded today dies in `_validate_header` with `Missing columns: Total Amount, Total Shares Amount`:

- `Total Amount` -> `Total Amount in Account Currency`
- `Total Shares Amount` -> `Total Amount in Instrument Currency`

Both are aliased back onto the names the parser already uses while the header is read, so `FreetradeTransaction` is untouched and old exports keep parsing exactly as before. The new tests use the header pasted in the issue verbatim.

Two judgement calls, happy to flip either:

- The `Stock Split *` columns are tolerated by prefix, not by listing all 15 names. The parser reads none of them, and an explicit list means the next column Freetrade adds reopens this issue. The cost is that a genuinely new split column won't be flagged.
- A missing column is still reported under its canonical name: an export lacking `Total Amount in Instrument Currency` reports `Missing columns: Total Shares Amount`. There's a test pinning that. Say the word if you'd rather it echoed the name the user sees in their file.

Out of scope: stock split rows themselves. A `STOCK_SPLIT` row still fails with `Unknown type: 'STOCK_SPLIT'`, and since parsing isn't year-filtered, one historical split anywhere in the file still blocks a run. Mapping Freetrade's split fields onto `ActionType.STOCK_SPLIT` (which wants the *additional* share count, while Freetrade gives "New Share Quantity" plus outturn rates) deserves its own PR rather than a guess that silently skews the multiplier. Happy to file a follow-up issue.

Checks: `uv run pytest` 282 passed, plus ruff, mypy and pylint (10/10).

Disclosure: written with AI assistance (Claude). The failure was reproduced against the header from #800 before anything was changed, and the new tests were confirmed to fail on the unfixed parser.

### DIFF
--- a/cgt_calc/parsers/freetrade.py
+++ b/cgt_calc/parsers/freetrade.py
@@ -62,6 +62,17 @@ class FreetradeColumn(StrEnum):
 COLUMNS: Final[list[str]] = [column.value for column in FreetradeColumn]
 REQUIRED_COLUMNS: Final[set[str]] = set(COLUMNS)
 
+# Freetrade renamed these columns, older exports still use the names above.
+COLUMN_ALIASES: Final[dict[str, str]] = {
+    "Total Amount in Account Currency": FreetradeColumn.TOTAL_AMOUNT.value,
+    "Total Amount in Instrument Currency": FreetradeColumn.TOTAL_SHARES_AMOUNT.value,
+}
+
+# Newer exports add a block of stock split columns. They are tolerated so that
+# the header validates, their values are not read yet: a stock split row still
+# fails as an unknown type.
+STOCK_SPLIT_PREFIX: Final = "Stock Split "
+
 
 def _parse_decimal(row: dict[str, str], column: FreetradeColumn) -> Decimal:
     """Parse Decimal value for column, raising ValueError with context on failure."""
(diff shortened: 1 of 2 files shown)


vmartinv:
Thanks! Feel free to add support for stock splits in another PR, this is already supported in other brokers I think.

PR #876, opened 2026-08-18
feat: NVIDIA support for the Schwab Equity Award JSON parser

ViieeS:
## Problem

The parser normalises splits for GOOG only. NVDA needs it too, and Schwab
treats it differently: acquisitions are restated into current units, disposals
are not. Missing it moves no money, only the share count, so nothing fails to
balance and the pool drifts from the broker every year after. On my own export,
54 shares across two 2023 disposals.

## Change

- NVDA disposals convert from the split dates, not a price heuristic — a
  heuristic needs the day's market price and expires once the real price passes
  it.
- A disposal's price comes from the money rather than the per-lot `SalePrice`,
  so a sale can span lots priced differently. Its quantity comes from the lots,
  or from the money where a lot stating zero shares hides the fraction from
  both count fields.
- Two checks raise: the split table against a `Lapse`, which states both units
  at once, and ESPP share accounting. A third warns — a disposal priced far
  below its neighbours has usually been restated twice, but a share that simply
  fell that far looks identical.
- Split handling moves into a per-symbol table. What differs between issuers is
  which records the export restated, so that is what the table states. GOOG
  becomes data, including its $175 floor.
- An ESPP purchase settled entirely in shares acquires nothing, and is no
  longer offered as an acquisition of zero. Only that case: a zero arriving any
  other way is a malformed row and is still refused.

The checks and the table are one change: both rest on the same fact about which
records the export restates.

## Verification

318 tests, `ruff` and `mypy` clean. GOOG is byte-identical before and after the
table change on every existing fixture, and those fixtures exercise the
heuristic on both sides. NVDA is covered by a synthetic portfolio with real
dates and prices and invented share counts, and by a real 146-row export not
included here.

## Open questions

- Table in-tree, or user-supplied config? GOOG cannot be expressed as config —
  its floor comes from that share's own price history, and a wrong one would
  corrupt a pool with nothing to fail on.
- That floor stays a heuristic. The `Lapse` cross-check has no equivalent for a
  partially restated export.

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,75 +145,68 @@ possible._
 
 #### NVIDIA equity awards
 
-The JSON parser handles NVDA as well as GOOG/GOOGL. Schwab reports NVDA history
-in mixed units, and which fields are restated for the 4:1 split of 20 July 2021
-and the 10:1 of 10 June 2024 varies by record type.
+The JSON parser handles NVDA as well as GOOG/GOOGL. Schwab reports NVDA history in mixed units, and
+which fields are restated for the 4:1 split of 20 July 2021 and the 10:1 of 10 June 2024 varies by
+record type.
 
-| Record | Read as |
-|---|---|
-| `Deposit` / `RS` — vest | `Quantity` and `VestFairMarketValue` as recorded; both are already restated and agree |
-| `Deposit` / `ESPP` | `NetSharesDeposited` when tax was settled in shares, else `Quantity`, priced at `PurchaseFairMarketValue` |
-| `Lapse` | checked for consistency, then skipped |
-| `Sale` | multiplied by the split factor for the trade date, priced from `(Amount + fees) / Quantity` |
+| Record                  | Read as                                                                                                   |
+| ----------------------- | --------------------------------------------------------------------------------------------------------- |
+| `Deposit` / `RS` — vest | `Quantity` and `VestFairMarketValue` as recorded; both are already restated and agree                     |
+| `Deposit` / `ESPP`      | `NetSharesDeposited` when tax was settled in shares, else `Quantity`, priced at `PurchaseFairMarketValue` |
+| `Lapse`                 | checked for consistency, then skipped                                                                     |
+| `Sale`                  | multiplied by the split factor for the trade date, priced from `(Amount + fees) / Quantity`               |
 
 Three of these are worth spelling out.
 
-**ESPP cost basis is the market value, not the price paid.** The discount is
-taxed as employment income through payroll, so the shares enter the pool at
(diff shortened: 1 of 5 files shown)


cgt_calc/parsers/schwab_equity_award_json.py
```diff
@@ -195,6 +306,191 @@ def _is_integer(number: Decimal) -> bool:
...
+
+        ratio = transaction.price / nearest.price
+
+        # Between "these agree" and "this one is a multiplier too cheap", take
+        # whichever is nearer on a log scale: ratio < 1/sqrt(multiplier), or
+        # ratio^2 * multiplier < 1 without a square root. The two hypotheses
+        # are a factor of 4 apart at the very least, so the midpoint is not a
+        # threshold that needs tuning.
+        if ratio * ratio * multiplier < 1:
+            raise ParsingError(
```
KapJI: If stock is very volatile price between vest and split can falsely trigger this. E.g. NVDA stock :)

Probably better as a warning?

README.md
```diff
@@ -143,6 +143,71 @@ _Note: For historic reasons, it is possible to provide the Equity Awards history
 [parser file](cgt_calc/parsers/schwab_equity_award_json.py). Please use the CSV method above if
 possible._
 
+#### NVIDIA equity awards
```
KapJI, line 146: There is too much prose, can be trimmed a bit to be closer to other sections.

cgt_calc/parsers/schwab_equity_award_json.py
ViieeS: Agreed, and 4:1 is the reachable one: NVDA fell ~56% in twelve weeks in late 2018, and the nearest vest can be six weeks out. Warning now, ff0d53b.

Kept `ParsingError` for the other two. A `Lapse` contradicting the split table and an ESPP purchase that does not add up are identities within the file, so no price move can reach them. Falsifiable -> raise, heuristic -> warn.

README.md
ViieeS: Trimmed in a7217d0, 66 lines down to 41: the record table, the three rules that change a number, and one paragraph on what is checked. The reasoning behind each now lives only in the code, next to what it constrains.

ViieeS:
One extra commit you did not ask for, 60dd5c3: the ESPP identity compared the counts inside a row against its restated `Quantity` without the split multiplier, unlike the `Lapse` check beside it. Harmless today — Schwab only began settling ESPP tax in shares after the last NVDA split — but the next split would make a correct file fail to parse. Both now go through one factor.

KapJI:
Thanks!

PR #924, opened 2026-08-21
Support no gain/no loss transfers to a spouse

### LINKED ISSUE #510: Support for "gift to spouse" action type

KapJI:
Adds a `TRANSFER_TO_SPOUSE` action for shares given to a spouse or civil partner, treated as no gain / no loss under [TCGA 1992 s58](https://www.legislation.gov.uk/ukpga/1992/12/section/58). Fixes #510, supersedes #549.

- The transfer is identified against the transferor's own same-day and 30-day acquisitions and the Section 104 pool exactly like a disposal, but each matched tranche's deemed proceeds equal its allowable cost, so the gain is nil and it stays out of the taxable disposal totals.
- The text report and PDF show the base cost passing to the recipient and the pool left behind, and transfers made in the reported period are logged. Price and fees on a transfer row cannot change the result, so they are ignored with a warning rather than silently.
- Refuses a sale and a transfer of the same symbol on the same day when an acquisition falls that day or in the next 30: [s105(1)](https://www.legislation.gov.uk/ukpga/1992/12/section/105) makes same-day disposals a single transaction, but one part is chargeable and the other is not, and nothing says how to split the matched acquisitions between them. Without such an acquisition both take the same pool average, so it calculates.
- Parses the Schwab Equity Awards `Gift` action instead of dying on an unknown label, and refuses until the recipient is known, printing the exact `TRANSFER_TO_SPOUSE` line to add. A matching row on the same date, symbol and quantity accounts for the broker's row, which is then dropped. This takes the useful half of #539; a gift to anyone else is a disposal at market value and is still unsupported.

No broker export marks these transfers, so they are added by hand in a RAW file, which can be passed alongside an existing broker file rather than replacing it.


cgt_calc/parsers/schwab_equity_award_json.py
```diff
@@ -833,8 +852,24 @@ def _convert_row_priced_before_a_split(self, history: SplitHistory) -> None:
...
+            # multiplier. Refuse rather than pick one.
+            raise ParsingError(
+                file,
+                f"Cannot tell how many shares the {self.symbol} gift of "
+                f"{self.date} moved. {self.symbol} split {multiplier}:1 after "
+                "that date, this export restates only some records, and a gift "
+                "has no price to tell them apart, so the count could be "
+                f"{strip_zeros(printed)} or "
+                f"{strip_zeros(printed * multiplier)}. "
+                "Check a statement and work this symbol out by hand.",
```
vmartinv: Can we also provide a way to avoid getting the error after working out the correct multiplier? Also does this happen in the csv format too? I never had a split there on my account.

cgt_calc/resources/template.tex.j2
```diff
@@ -97,9 +100,21 @@
...
+    \BLOCK{ set pool = entries[-1] }
+    Number of units left in the pool: \VAR{ strip_zeros(pool.new_quantity) }%
+    \BLOCK{ if pool.new_quantity > 0 },
+        new pool cost: £\VAR{ "{:,}".format(round_decimal(pool.new_pool_cost, 2)) }
+        (£\VAR{ "{:,}".format(round_decimal(pool.new_pool_cost/pool.new_quantity, 2)) }/unit)%
+    \BLOCK{ endif }.
 \BLOCK{ endif }
 
-\BLOCK{ if not is_rename }
+\BLOCK{ if not is_rename and not is_transfer_to_spouse }
```
vmartinv, line 117: can you add an e2e test that generates the pdf with it, then we can visually check it from the CI step. It can also validate the stdout/stderr output.

It would be good to have 2 test reports: one for the sender and one for the recipient of the shares.

tests/general/test_calc.py
vmartinv, line 68: side comment, test are more clear if it mimics the tool's args behaviour:
- the requested tax year is explicit
- the default is to check the balance 


```suggestion
    tax_year: int, balance_check: bool = True
```

might be better to do it in a separate PR

tests/general/test_calc.py
```diff
@@ -933,6 +939,846 @@ def test_disposal_debug_log_keeps_fractional_quantity(
     assert all("quantity 2.5" in message for message in disposal_logs)
 
 
+def test_transfer_to_spouse_from_pool_is_no_gain_no_loss() -> None:
```
vmartinv: Since they are so many, the transfer to spouse tests should be in a separate file

vmartinv (changes requested):

tests/general/test_calc.py
KapJI: Done

cgt_calc/parsers/schwab_equity_award_json.py
KapJI: Good catch, fixed. No error at parse time now, it prints both possible lines and you add whichever your statement confirms. That row settles the count and the recipient in one go.

CSV is unaffected. It has no split restatement logic at all, and no `Gift` action either, so a gift there fails as an unknown action long before this could come up.

cgt_calc/resources/template.tex.j2
KapJI: Added

docs/brokers/raw.md
```diff
@@ -23,3 +23,42 @@ Example usage for the tax year 2024/25:
...
+No broker export marks these, so you add the row yourself. You do not have to convert your whole
+history to the RAW format to do it: pass a small RAW file of just the transfers alongside your
+broker export.
+
+```shell
+cgt-calc --year 2024 --schwab-file transactions.csv --raw-file transfers.csv
+```
+
+The report and the PDF show the base cost that passes to the recipient. Pass that figure on to them:
+it is what they will need for their own return when they come to sell.
```
vmartinv: We should clarify what the spouse needs to do for the report as well. I see in the e2e test that it adds a cash transfer for the cost + a buy transaction. I'm not sure how the gift shows up in the actual broker. I'm assuming both of them will be using this tool.

vmartinv (approved):

docs/brokers/raw.md
KapJI: Good point, thanks. Added TRANSFER_FROM_SPOUSE for the other side: quantity and base cost per unit, no cash, pooled like a vest. The transferor's report now prints the exact row to hand over.

On how the gift shows up at the broker: honestly no idea. Whatever they show, they won't know the cost, so the RAW row is the answer either way.

tests/general/test_calc.py
KapJI: Agreed, will do it in a separate PR

PR #988, opened 2026-08-24
Skip ISINs with no known ticker when building the symbol map

amaksoft:
An ISIN whose ticker is not known is stored with an empty symbol, which validate_data explicitly permits, and initial_isin_translation.csv ships one. get_symbol_to_isin_map passed that straight through, so the map gained an "" key pointing at whichever ISIN happened to hold it.

Callers resolve a transaction's ISIN with isin_map.get(trx.symbol or ""), so every transaction that legitimately has no symbol - credit interest, wire transfers, service fees - matched that arbitrary ISIN. The visible effect was ERI rows for a fund the user has never held being pulled into the transaction set.

### DIFF
--- a/cgt_calc/isin_converter.py
+++ b/cgt_calc/isin_converter.py
@@ -128,11 +128,18 @@ class IsinConverter:
         return result
 
     def get_symbol_to_isin_map(self) -> dict[str, Isin]:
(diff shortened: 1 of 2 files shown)


KapJI:
Thanks!

PR #989, opened 2026-08-24
Add iShares IV reportable income for the period to 31 May 2025

amaksoft:
Imported from the published investor report with
scripts/import_eri_reports.py, which recognised it as a Blackrock/iShares file. Covers the 174 share classes in iShares IV plc whose reporting period ended 31 May 2025, deemed distributed on 30 November 2025 and so taxable in 2025/26.

The bundled data ran to 31 May 2024 for this umbrella, so every fund with that year end reported nil for the year regardless of its actual figure.

Source: https://www.ishares.com/gls-download/literature/tax-information/isharesivplc31052025-investorreport.xlsx


KapJI:
Thanks!

PR #991, opened 2026-08-24
Extract a stock split handler and merge two identical branches

amaksoft:
No behaviour change. convert_to_hmrc_transactions sits on ruff's max-complexity limit, so any new action cannot be added to it without tripping C901 in the same commit that adds the feature.

Moves the stock split body into record_stock_split alongside the other per-action helpers, and merges the ADJUSTMENT and WIRE_FUNDS_RECEIVED branches, which were the same three lines.


KapJI:
Thanks!

PR #995, opened 2026-08-24
Refuse a Cancel Buy that has no matching Buy

amaksoft:
Schwab emits a "Cancel Buy" row to reverse a purchase it has already reported, and the parser paired the two so both drop out. Two problems sat in that pairing.

The search for the Buy compared action types. A Cancel Buy was itself mapped to ActionType.BUY, so one cancellation could pair with another and leave both of the purchases they reversed standing. Match on the raw action instead, and mark each Buy as it is consumed so two cancellations cannot claim the same one.

When no Buy was found the parser warned and carried on. The row stayed typed as a purchase, so a reversal silently became cost basis and every later disposal of that symbol was computed against a pool that never existed. A warning is the wrong tool for a wrong number: raise ParsingError naming the row instead.

Give Cancel Buy its own ActionType.CANCEL_BUY as well. Modelling it as a purchase left correctness resting on a filter running first, and made the type say the opposite of what the row means.

### DIFF
--- a/.harper-dictionary.txt
+++ b/.harper-dictionary.txt
@@ -101,6 +101,7 @@ NVDA
 NVIDIA
 OPRA
 OTGLY
+ParsingError
 pdf
 pdflatex
 pipx
--- a/.harper-dictionary.txt
+++ b/.harper-dictionary.txt
@@ -72,6 +72,7 @@ HL
 Hmrc
 IBKR
 IBKR's
+idx
 InvalidTransactionError
 Invesco
 InvestmentName
@@ -161,6 +162,7 @@ tranche's
 TransactionDetails
 transferor
 transferor's
+txn
 ty
 UCITS
 un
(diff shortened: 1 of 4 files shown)


KapJI:
Thanks! Rebased onto main and pushed three commits.

A cancellation and the purchase it reverses can share a date, and the export gives no order to rows within a date, so the purchase is sometimes listed above the cancellation. The forward-only search refused those files. Same-date rows above the cancellation are now searched too — only same-date ones, since a later Buy had not happened when the cancellation was recorded and consuming it would delete a real acquisition.

Added the regression test for the pairing bug. `test_multiple_cancel_buys_same_symbol` varies quantity and price between the two pairs, so it passes on the old code as well; two identical Cancel Buy rows against two identical Buy rows is what reproduces it.

The error message now names all four causes and qualifies the remedy. Given two cancellations and one purchase, the newer takes it and the older raises — so removing that row *and* the purchase it names breaks the pair that did reconcile. Removing the cancellation alone is right there.

I left the search window as a constant rather than a flag. The only reason to widen it is to clear the error, and that is the direction that lets a cancellation match an unrelated purchase and silently delete a real acquisition.

One note for the description: the guard stopping two cancellations claiming the same Buy predates this branch, so the pairing fix is the `raw_action` check on its own.

PR #1000, opened 2026-08-25
Support the current Trading 212 export and read dates in UK time

casperdcl:
Haven't bothered updating tests, but seems like the format needs updating (at least as per exports today).


cgt_calc/parsers/trading212.py
```diff
@@ -236,8 +236,7 @@ def __init__(self, header: list[str], row_raw: list[str], file: Path) -> None:
         }
 
         time_str = row[Trading212Column.TIME]
-        time_format = "%Y-%m-%d %H:%M:%S.%f" if "." in time_str else "%Y-%m-%d %H:%M:%S"
-        self.datetime = datetime.strptime(time_str, time_format)
+        self.datetime = datetime.strptime(time_str[:19], "%Y-%m-%d %H:%M:%S")
```
casperdcl: New format has an extra `+00:00` at the end. Seems pointless to parse too many formats explicitly given that ultimate resolution is in days anyway. But perhaps should do...

```suggestion
        time_format = (
            "%Y-%m-%d %H:%M:%S%z"
            if "+" in time_str
            else "%Y-%m-%d %H:%M:%S.%f"
            if "." in time_str
            else "%Y-%m-%d %H:%M:%S"
        )
        self.datetime = datetime.strptime(time_str, time_format)
```

cgt_calc/parsers/trading212.py
```diff
@@ -304,9 +303,16 @@ def __init__(self, header: list[str], row_raw: list[str], file: Path) -> None:
...
-                check_price = abs(amount + check_fees) / quantity
-                calculated_price_foreign = check_price * exchange_rate
-                discrepancy = self.price_foreign - calculated_price_foreign
+                if action == ActionType.DIVIDEND:
+                    discrepancy = (
+                        self.price_foreign * quantity * exchange_rate
+                        - amount
+                        - check_fees
+                    )
+                else:
```
casperdcl: technically not required to "fix" but was nice to suppress warnings. Haven't double-checked the formula.

KapJI:
Thanks! I can review it and fix tests later.

Just to confirm: these changes are coming from your recent export?

In the meantime, please take a look at https://github.com/tqdm/shtab/pull/256 🙂

casperdcl:
Tested manually today, yes.

Also, while we're off-topic: https://github.com/KapJI/capital-gains-calculator/blob/0adf60ce8f83600c6cf8ef5f90564724fccd78fc/cgt_calc/isin_converter.py#L106 was annoying. Can it be a warning instead of an exception? Or perhaps at least have some common aliases built in to suppress the error?

```csv
ISIN,symbol
US67066G1040,NVD,NVDA
US11135F1012,1YD,AVGO
```

vmartinv (dismissed):
it looks like this change would break the previous format. Can you make the change backwards compatible?

cgt_calc/parsers/trading212.py
```diff
@@ -160,6 +161,23 @@ def decimal_or_none(
...
+    converted to UTC and made naive so every transaction stays comparable.
+    """
+
+    try:
+        parsed = datetime.fromisoformat(value)
+    except ValueError as err:
+        raise ValueError(f"Invalid timestamp: {value!r}") from err
+    if parsed.tzinfo is None:
+        return parsed
+    return parsed.astimezone(UTC).replace(tzinfo=None)
```
casperdcl: given that the tax year end falls during BST, might be better to use (or make this configurable):

```suggestion
    return parsed.astimezone(timezone(timedelta(hours=1))).replace(tzinfo=None)
```

KapJI: Probably should be `Europe/London` but it's a separate issue.

Edit: Actually yes, let's fix T212 part here.

PR #1010, opened 2026-08-25
Read Freetrade dates in UK time

### LINKED ISSUE #1005: Freetrade dates use the UTC calendar day instead of the UK one
cgt_calc/parsers/freetrade.py:196 takes `.date()` straight from the UTC timestamp. UK tax dates are UK calendar days, and 5 April always falls inside BST, so a transaction stamped after 23:00 UTC on 5 April is filed in the previous tax year. It can also affect same-day and 30-day matching.

Trading 212 was fixed in #1000; `UK_TIMEZONE` is now in `const.py`, so the fix is to convert before taking the date, with a 5/6 April boundary test.

Separately, the date-only parsers (Schwab, Interactive Brokers, Hargreaves Lansdown, Vanguard, MSSB, Sharesight) export a broker-local date with no time, so their dates cannot be corrected. This should be stated in the docs.

hdimer:
Fixes #1005.

`FreetradeTransaction.__init__` took the calendar date straight from the UTC timestamp, so a transaction stamped after 23:00 UTC on 5 April was filed in the previous tax year, and the same instant could also land on the wrong side of same-day and 30-day matching.

It now converts to `UK_TIMEZONE` before taking the date, the same way `Trading212Transaction` has since #1000. As there, a zone-less timestamp is read as UTC rather than as the machine's local time: every export carries `Z`, but without that the parsed date would depend on the reader's clock, and `BASE_ROW_VALUES` in the Freetrade tests uses the zone-less shorthand.

The new parametrized test mirrors the Trading 212 one: the 5/6 April boundary (22:59:59 UTC still 5 April, 23:00:00 and 23:30:00 already 6 April, the last of those zone-less), the evening after the clocks go back in October, and two winter dates where UK and UTC agree anyway. I checked each case discriminates by re-running it against the unfixed parser (two boundary cases fail), against a plain `astimezone` with no zone-less handling under `TZ=Pacific/Auckland` (the zone-less case fails), and against a crude "add an hour from April to October" version (the October case fails).

The second half of the issue is documented rather than fixed. `docs/brokers/freetrade.md` gains a "Dates and time zones" section mirroring the Trading 212 one, and `docs/brokers/index.md` says once that every other export states a date with no time, so cgt-calc reads it as the broker wrote it. That seemed better than repeating the same paragraph on six broker pages, but happy to move it if you would rather each page carried it.

Checks: `uv run pytest` 904 passed, and the Freetrade tests also pass under `TZ` set to UTC, Auckland, Los Angeles and Tokyo. Plus ruff format, ruff check, mypy, ty, dprint, codespell and Harper.

Disclosure: written with AI assistance (Claude). The bug was reproduced against `main` first, and the new test was confirmed to fail on the unfixed parser before anything was changed.

### DIFF
--- a/cgt_calc/parsers/freetrade.py
+++ b/cgt_calc/parsers/freetrade.py
@@ -3,12 +3,13 @@
 from __future__ import annotations
 
 import csv
-from datetime import datetime
+from datetime import UTC, datetime
 from decimal import Decimal, InvalidOperation
 from enum import StrEnum
 import logging
 from typing import TYPE_CHECKING, ClassVar, Final, TextIO, override
 
+from cgt_calc.const import UK_TIMEZONE
 from cgt_calc.exceptions import (
     ParsingError,
     UnsupportedBrokerActionError,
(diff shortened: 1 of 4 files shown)


vmartinv:
Thanks!

PR #1012, opened 2026-08-26
Add Schwab written equity option support

moguzozcan:
Schwab exports a written option as separate rows: the grant, whatever ended it, and for an
assignment a stock row of its own. cgt-calc could not read any of them.

- Recognise `Sell to Open`, `Buy to Close`, `Expired` and `Assigned`, both Schwab symbol formats,
  and the accounting parentheses it writes a debit in.
- Treat the grant as the disposal (TCGA 1992 s144): a lapsed option keeps the premium as its gain,
  a closing debit is an allowable cost of the grant converted on its own date, and an assigned
  premium moves into the underlying share transaction.
- Date an assigned share transaction on the assignment rather than on Schwab's settlement row,
  which can fall in the next tax year.
- Refuse rather than guess: a settlement row that is missing, duplicated or does not reconcile with
  the strike; a contract assigned twice in a day; an adjusted or cash-settled root; a malformed
  premium row.
- Reconcile once per account, so an option opened and closed in different files of one
  `--schwab-dir` export still resolves.

Purchased options and box spreads are follow-ups, so nothing here classifies a box or reports
financing. `_record_interest` is extracted from `convert_to_hmrc_transactions` because the added
dispatch branch took it past the complexity limit, and a suppression is not the fix.


cgt_calc/model.py
```diff
@@ -206,6 +206,142 @@ def __add__(self, amount: ForeignCurrencyAmount) -> ForeignCurrencyAmount:
...
+    option_type: OptionType
+    multiplier: Decimal = Decimal(100)
+
+    @override
+    def __str__(self) -> str:
+        """Return a compact, human-readable contract name."""
+        return (
+            f"{self.underlying} {self.expiry.isoformat()} "
+            f"{strip_zeros(self.strike)} {self.option_type.value}"
+        )
```
vmartinv: This doesn't seem human readable, can you add connectors (expiring, for, etc) and something that indicates it is an option?

cgt_calc/parsers/schwab.py
```diff
@@ -664,6 +775,890 @@ def _filter_cancelled_buy_transactions(
...
+        quantity=None,
+        price=None,
+        fees=Decimal(0),
+        amount=amount * proportion,
+        currency=transaction.currency,
+        broker=transaction.broker,
+    )
+
+
+def _reconcile_written_options(  # noqa: C901
```
vmartinv: reconcile_written_options and some parsing functions are very long (complex). Consider splitting into smaller functions or helper classes for clarity and testability.

cgt_calc/parsers/schwab.py
```diff
@@ -664,6 +775,890 @@ def _filter_cancelled_buy_transactions(
...
+    call_is_purchase = holder
+    expected_action = (
+        ActionType.BUY
+        if ((contract.option_type is OptionType.CALL) == call_is_purchase)
+        else ActionType.SELL
+    )
+    candidates = [
+        transaction
+        for transaction in transactions
+        if id(transaction) not in used
```
vmartinv: this is risky giving we didnt define it for Schwab transactions

cgt_calc/main.py
```diff
@@ -1021,7 +1136,7 @@ def _convert_foreign_fees(self, transaction: BrokerTransaction) -> None:
                 abs(transaction.amount + transaction.fees) / transaction.quantity
             )
 
-    def convert_to_hmrc_transactions(
+    def convert_to_hmrc_transactions(  # noqa: C901
```
vmartinv: consider splitting the function
```suggestion
    def convert_to_hmrc_transactions(
```

cgt_calc/resources/template.tex.j2
```diff
@@ -15,6 +15,29 @@
 \BLOCK{ set interest_tax_count = namespace(value=0) }
 \BLOCK{ set eri_count = namespace(value=0) }
 
+\BLOCK{ if report.box_spreads }
```
vmartinv: We need to add an E2E test that generates the PDF with these transactions so we can verify they are being rendered correctly.

vmartinv (dismissed):
Left some comments. Due to the complexity of this feature I suggest we first only add option support in this PR and then create a follow up for box spreads, this way review becomes more manageable.

cgt_calc/model.py
KapJI: Done - it now reads `META call option expiring 2024-05-17 at a strike of 500`, and there is a test for it.

cgt_calc/parsers/schwab.py
KapJI: Moved all of this into `schwab_options.py` and split it into a scan pass and an emit pass, so the long functions are gone.

cgt_calc/parsers/schwab.py
KapJI: Good catch - rows are keyed on the file and line from `TransactionSource` now, instead of `id(transaction)`.

cgt_calc/main.py
KapJI: Split it up - the grant and close work is in helpers now, so the dispatcher only gets one extra branch.

cgt_calc/resources/template.tex.j2
KapJI: Added one - it runs the CLI over a small export and checks the report, and builds the PDF in the `ENABLE_PDFLATEX=1` job.

PR #1013, opened 2026-08-26
Reject '-' for the cache file options

### LINKED ISSUE #1009: Cache file options accept `-` and write a file literally named `-`
## Problem

`optional_file_type` maps `-` to the stdin sentinel:

https://github.com/KapJI/capital-gains-calculator/blob/e796208d0ac988648b7501fd1a8b53e06d51ee0b/cgt_calc/args_validators.py#L85-L91

That is correct for input files, but the three options using it are read/write caches, where stdin has no meaning:

- `--exchange-rates-file`
- `--isin-translation-file`
- `--spin-offs-file`

https://github.com/KapJI/capital-gains-calculator/blob/e796208d0ac988648b7501fd1a8b53e06d51ee0b/cgt_calc/args_parser.py#L105-L133

Nothing downstream guards the sentinel, so every writer treats `Path("-")` as an ordinary relative path:

https://github.com/KapJI/capital-gains-calculator/blob/e796208d0ac988648b7501fd1a8b53e06d51ee0b/cgt_calc/isin_converter.py#L187-L196

https://github.com/KapJI/capital-gains-calculator/blob/e796208d0ac988648b7501fd1a8b53e06d51ee0b/cgt_calc/spin_off_handler.py#L53-L59

https://github.com/KapJI/capital-gains-calculator/blob/e796208d0ac988648b7501fd1a8b53e06d51ee0b/cgt_calc/currency_converter.py#L183-L191

Passing `-` therefore creates a file named `-` in the working directory rather than doing anything with stdin. Two follow-on effects:

- On the next run `Path("-").is_file()` is true, so the file is also read back. The cache has silently relocated to a path the user never asked for, and the real cache is neither read nor updated.
- A file named `-` is awkward to remove, needing `rm -- -` or `rm ./-`.

## Reproduction

In an empty directory, with the package installed:

```python
from pathlib import Path

from cgt_calc.currency_converter import CurrencyConverter
from cgt_calc.isin_converter import IsinConverter
from cgt_calc.model import Isin
from cgt_calc.spin_off_handler import SpinOffHandler

converter = IsinConverter(isin_translation_file=Path("-"))
converter.write_data[Isin("IE00BK5BQT80")] = {"VWRP"}
converter._write_isin_translation_file()
print(Path("-").read_text())  # ISIN,symbol / IE00BK5BQT80,VWRP

handler = SpinOffHandler(spin_offs_file=Path("-"))
handler.cache = {"NEW": "OLD"}
handler._write_spin_off_file()
print(Path("-").read_text())  # dst,src / NEW,OLD

CurrencyConverter._write_exchange_rates_file(Path("-"), {})
print(Path("-").read_text())  # month,currency,rate
```

Each call leaves a file literally named `-` in the working directory. Note the ISIN case only writes under `RuntimeMode.PROD`, so reproduce with `CGT_TEST_MODE` and `CGT_TEST_MODE_STRICT` unset.

## Proposed behavior

Reject `-` for these three options, since a read/write cache cannot be stdin. A dedicated validator raising `argparse.ArgumentTypeError` keeps the failure at parse time, next to the existing "expected file path, got directory" message, rather than surfacing as a stray file after the run.

An empty value already maps to `None` and correctly disables the cache, so `-` is the only unhandled case.

This would also let the ERI warning added in #1002 drop its `STDIN_PATH` special-case: it currently has to avoid naming `-` as a place to add mappings, which encodes the same "`-` is not a real cache" knowledge in a second place.

hdimer:
Fixes #1009.

`optional_file_type` mapped `-` to the stdin sentinel, but all three of its call sites are the read/write caches, where stdin has no meaning. Nothing downstream guarded the sentinel, so `-` created a file literally named `-` in the working directory, and the next run read that file back as the cache while the intended one was never touched.

Since every caller was a cache option, I narrowed that validator in place and renamed it `cache_file_type`, rather than adding a second one and leaving `optional_file_type` with no callers. Happy to split it into two if you would rather keep the general form. An empty value still disables the cache, and `-` still means stdin for the input options through `existing_file_type`.

The guard matches the argument exactly, so `./-` still resolves to `Path("-")` and still writes a file named `-`. That felt like the right line to draw: it is the conventional unambiguous way to name a file called `-`, it is deliberate rather than a surprise, and nothing compares these three options against the sentinel anymore. Say the word if you would rather reject the resolved path instead.

The second half of the issue follows from that: `broker_registry` no longer needs its `STDIN_PATH` check, because `--isin-translation-file` can no longer hold the sentinel. That also leaves `test_vanguard_warning_omits_stdin_sentinel_as_cache_path` asserting something untrue, since `location` would now render `" at -"`, so I removed it rather than reworking it to build the sentinel by hand and pin behaviour the CLI can no longer produce. Both branches of the simplified conditional stay covered by the two neighbouring tests.

The new parser test covers all three options and asserts the parse-time message. I checked it discriminates by running it against the unfixed validator, where all three cases fail with `DID NOT RAISE SystemExit`, and added a companion case for `--exchange-rates-file=-rates.csv` because an over-broad `startswith("-")` guard would otherwise pass the whole suite. While I was in the file I gave the four neighbouring tests over the same three options the same `cache_path_arguments` name, since they all exercise this validator.

Reproducing on `main` first also showed the read-back you describe: writing the ISIN cache to `-` and then constructing `SpinOffHandler` with `-` made it load the ISIN file as its own cache and fail on `invalid columns dict_keys(['ISIN', 'symbol'])`.

Left out of scope, mention it if you want it: `output_path_type` also accepts `-`, so `-o -` writes the report to a file named `-`. That one is write-only and never read back as data, so it is a different bug from the cache round-trip this issue is about.

Checks: `uv run pytest` 910 passed, with and without `CGT_TEST_MODE_STRICT`. Plus ruff format, ruff check, mypy, ty, codespell and Harper.

Disclosure: written with AI assistance (Claude). The bug was reproduced against `main` first, and the new test was confirmed to fail on the unfixed validator before anything was changed.

### DIFF
--- a/cgt_calc/args_validators.py
+++ b/cgt_calc/args_validators.py
@@ -82,12 +82,17 @@ def _ensure_readable_directory(path: Path, value: str) -> None:
         ) from err
 
 
-def optional_file_type(value: str) -> Path | None:
-    """Convert non-empty value to Path and ensure file semantics."""
+def optional_cache_file_type(value: str) -> Path | None:
+    """Convert non-empty value to a cache Path and ensure file semantics.
+
+    An empty value disables the cache. The stdin marker is rejected: a cache is
+    read and written, so '-' would create a file literally named '-' and later
+    be read back as the cache.
+    """
     if value.strip() == "":
         return None
     if value == "-":
-        return STDIN_PATH
+        raise argparse.ArgumentTypeError("expected file path, got stdin marker: '-'")
     path = Path(value)
     if path.exists():
         if not (path.is_file() or path.is_fifo()):
--- a/cgt_calc/args_parser.py
+++ b/cgt_calc/args_parser.py
@@ -13,7 +13,7 @@ from .args_validators import (
     VersionAction,
     date_type,
     existing_file_type,
-    optional_file_type,
+    optional_cache_file_type,
     output_path_type,
     set_completer,
     ticker_list_type,
@@ -105,7 +105,7 @@ Environment variables:
     set_completer(
         data_group.add_argument(
             "--exchange-rates-file",
-            type=optional_file_type,
+            type=optional_cache_file_type,
             metavar="PATH",
             default=DEFAULT_EXCHANGE_RATES_FILE,
             help="monthly exchange rates in CSV format (generated automatically if missing; default: %(default)s)",
@@ -115,7 +115,7 @@ Environment variables:
     set_completer(
         data_group.add_argument(
             "--isin-translation-file",
-            type=optional_file_type,
+            type=optional_cache_file_type,
             default=DEFAULT_ISIN_TRANSLATION_FILE,
             metavar="PATH",
             help="ISIN to ticker translations in CSV format (generated automatically if missing; default: %(default)s)",
@@ -125,7 +125,7 @@ Environment variables:
     set_completer(
         data_group.add_argument(
             "--spin-offs-file",
-            type=optional_file_type,
+            type=optional_cache_file_type,
             metavar="PATH",
             default=DEFAULT_SPIN_OFF_FILE,
             help="spin-offs data in CSV format (default: %(default)s)",
--- a/cgt_calc/parsers/broker_registry.py
+++ b/cgt_calc/parsers/broker_registry.py
@@ -8,7 +8,6 @@ from typing import TYPE_CHECKING, ClassVar
 
 from colorama import Fore
 
-from cgt_calc.args_validators import STDIN_PATH
 from cgt_calc.const import RENAME_DESCRIPTION_PREFIX
 from cgt_calc.logging import style_text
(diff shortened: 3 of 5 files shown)


cgt_calc/args_validators.py
```diff
@@ -82,12 +82,17 @@ def _ensure_readable_directory(path: Path, value: str) -> None:
         ) from err
 
 
-def optional_file_type(value: str) -> Path | None:
-    """Convert non-empty value to Path and ensure file semantics."""
+def cache_file_type(value: str) -> Path | None:
```
vmartinv:
```suggestion
def optional_cache_file_type(value: str) -> Path | None:
```

vmartinv (dismissed):
just a minor suggestion

PR #1018, opened 2026-08-26
Trading212 Stock split open/close

casperdcl:
Haven't added tests, but works locally with recent CSV export.

Added a `# type: ingore` in lieu of new `ActionType`s.

### DIFF
--- a/cgt_calc/parsers/trading212.py
+++ b/cgt_calc/parsers/trading212.py
@@ -72,6 +72,8 @@ class Trading212Column(StrEnum):
     CURRENCY_FINRA_FEE = "Currency (Finra fee)"
     MERCHANT_CATEGORY = "Merchant category"
     MERCHANT_NAME = "Merchant name"
+    STOCK_SPLIT_OPEN = "Stock split open"
+    STOCK_SPLIT_CLOSE = "Stock split close"
 
 
 COLUMNS: Final[list[str]] = [column.value for column in Trading212Column]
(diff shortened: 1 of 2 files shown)


vmartinv (changes requested):
Could you add a test for this change? it's unclear what split open/close means here

KapJI:
Just adding tests isn't enough. Proper support for this is more complicated. I will try to implement better support for these transactions.

casperdcl:
Could you give [a few pointers](https://xkcd.com/138/) on what some of the problems are? I imagine there are some side-effects on summary calculations?

KapJI:
The reason it needed the calculator side as well:

- **Every consolidation fails.** For a reverse split `open - close` is negative, and `add_stock_split` passes that to `add_acquisition`, which raises `QuantityNotPositiveError`. Reverse splits are most of what shows up in the sanitised exports I could find (XXII, AVX, ULVR, RKT).
- **Row order.** Pairing on `transactions[-1]` assumes open-then-close, but four of the seven real pairs I found are written close-first (BYDDY, NFLX, ULVR, XXII).
- **Halves in different exports.** Exports are cut by date range and `read_transactions` runs per file, so the two rows can land either side of the seam and never see each other.
- **No ratio.** A net share delta can't convert counts across the event, so a 30-day match after a split (or a trade earlier on the split day) uses the wrong number of shares.
- **Multiple brokers.** The Section 104 pool is global. Hold the same stock at two brokers and it's one holding. So one account's share delta can't be applied to it: `(delta + holding) / holding` gives 1.5x for a 2-for-1 when half the holding is elsewhere, and the pool silently ends up short.

PR #1027, opened 2026-08-27
Add Revolut GIA

casperdcl:
- add Revolut & docs
- ~had to make balance check against `-0.01` rather than `0` due to Revolut allowing negative balances~
- ~haven't added any tests~


cgt_calc/parsers/revolut.py
```diff
@@ -0,0 +1,266 @@
...
+            quantity,
+            price,
+            Decimal(0),
+            amount,
+            currency,
+            "Revolut",
+        )
+
+
+class RevolutParser(BaseSingleFileParser):
```
vmartinv: this can use the standard csv parser
```suggestion
class RevolutParser( StandardCSVParser):
```

cgt_calc/main.py
```diff
@@ -1139,7 +1139,7 @@ def convert_to_hmrc_transactions(
                     transaction, f"Action not processed({transaction.action})"
                 )
             balance_history.append(new_balance)
-            if self.balance_check and new_balance < 0:
+            if self.balance_check and new_balance < -Decimal("0.01"):
```
vmartinv: dont change this, please find the source of the discrepancy in the source itself

vmartinv (changes requested):
we will also need tests (including the e2e test), you can check the other brokers for reference

cgt_calc/main.py
casperdcl: revolut actually allows/allowed small negative balances. I'd suggest making this configurable (`--balance-check-min=-0.01` or similar).

vmartinv:
We are still missing the tests (unit and e2e)

cgt_calc/main.py
vmartinv: until that feature is available feel free to use `--no-balance-check`

vmartinv:
Thank you!

vmartinv (approved):

PR #1032, opened 2026-08-28
Add argument for excluding some tickers from CGT

0xr0bert:
This adds the argument:

--cgt-exempt-tickers TICKER[,TICKER]

which is then excluded for CGT purposes (but retained for other purposes like interest). This is useful for UK Gilts and other qualifying bonds.

### DIFF
--- a/cgt_calc/main.py
+++ b/cgt_calc/main.py
@@ -223,10 +223,9 @@ class CapitalGainsCalculator:
         self.balance_check = balance_check
         self.calc_unrealized_gains = calc_unrealized_gains
         self.interest_fund_tickers = interest_fund_tickers
-        self.cgt_exempt_tickers = [
-            ticker.upper() for ticker in (cgt_exempt_tickers or [])
-        ]
-        self.cgt_exempt_tickers_set = set(self.cgt_exempt_tickers)
+        self.cgt_exempt_tickers = frozenset(
+            ticker.upper() for ticker in cgt_exempt_tickers or []
+        )
         self.total_uk_interest = Decimal(0)
         self.total_foreign_interest = Decimal(0)
         self.total_interest_tax = Decimal(0)
@@ -312,7 +311,7 @@ class CapitalGainsCalculator:
 
     def is_cgt_exempt(self, symbol: str) -> bool:
         """Check if symbol is exempt from Capital Gains Tax."""
-        return symbol.upper() in self.cgt_exempt_tickers_set
+        return symbol.upper() in self.cgt_exempt_tickers
 
     def get_eri(self, symbol: str, date: datetime.date) -> ExcessReportedIncome | None:
         """Return Excess Reported Income at specific date for the input symbol."""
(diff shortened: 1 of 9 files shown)


cgt_calc/resources/template.tex.j2
```diff
@@ -48,6 +48,9 @@
     \BLOCK{ set symbol = symbol[17:] }
     \BLOCK{ set is_disposal = True }
     \BLOCK{ set is_gift = True }
+\BLOCK{ elif symbol.startswith("exempt$") }
```
vmartinv, line 51: you will need a test that generates the pdf so we can check it's being generated correctly

cgt_calc/main.py
```diff
@@ -2529,20 +2544,46 @@ def calculate_capital_gain(
...
+                            exempt_disposal_count += 1
+                            exempt_disposal_proceeds += transaction_disposal_proceeds
+                            LOGGER.debug(
+                                "EXEMPT DISPOSAL on %s of %s, quantity %s, proceeds £%s",
+                                date_index,
+                                symbol,
+                                transaction_quantity,
+                                round_decimal(transaction_disposal_proceeds, 2),
+                            )
+                            calculation_log[date_index][f"exempt${symbol}"] = (
```
vmartinv: add exempt into _disposal_log_prefix

```suggestion
                            calculation_log[date_index][f"{prefix}${symbol}"] = (
```

vmartinv (dismissed):
Please address the CI errors

vmartinv:
as a follow up, is there some source that provides a list of all exempt assets? preferably ISIN indexed. It would be much better if this was bundled with the tool, then they would get exempt automatically. As such it may be better to provide this list as a file rather an a list.

0xr0bert:
Changes made. There does not exist a list of all exempt assets.

There is one for gilts, but it is frequently updated as the DMO issues new gilts, but there does not exist one for QCBs.

vmartinv:
There is a list of UK gilts here https://www.dmo.gov.uk/data/pdfdatareport?reportCode=D1A

it would be nice to be able to parse the excel report .
We would still need a way to provide other exempt assets though, so this doesn’t seem bad.

0xr0bert:
Well it's a bit tricky as that data only includes currently live gilts, historical ones (or future ones) aren't included. Gilts are issued frequently - there have been four issued in August alone, so this would have to be updated frequently, and some historical record kept.

It also would not resolve the issue for QCBs of which there is no list.

KapJI:
Pushed a follow-up commit addressing the review:

- Reframed this as an advanced manual “CGT-exempt instrument classification” override, not general gilt/QCB support. The classification is the user’s unverified assertion; QCB status cannot be inferred from a ticker or name.
- Clarified that both gains and losses are disregarded, while coupon and accrued interest remain taxable income. The Accrued Income Scheme is not implemented.
- Added same-day, 30-day, income, exempt-gift, and PDF-generation regression coverage.
- Corrected the PDF wording and removed the misleading `--interest-fund-tickers` recommendation. The existing broker-guide caveats remain intact.

KapJI:
Thanks!

PR #1034, opened 2026-08-28
Parameterise the base parsers over their transaction type

KapJI:
`BaseSingleFileParser`, `StandardCSVParser` and `BaseDirParser` are now generic over the transaction type they produce, so each parser declares what it actually returns. `Trading212Parser._by_date_and_action` no longer needs `assert isinstance(transaction, Trading212Transaction)` — permanently unreachable, since `post_process_transactions` is only called with the parser's own rows — nor the `type: ignore[arg-type]` the assert replaced. Neither remains anywhere in `cgt_calc/parsers`.

Eight of the eleven parsers gain precise types: Raw, Sharesight, Schwab, Schwab equity awards, Vanguard, Interactive Brokers, ERI raw and Trading 212. Three cannot and are parameterised at `BrokerTransaction` for uniformity: Freetrade genuinely returns a mixed list (`FreetradeTransaction` rows plus bare `BrokerTransaction` rows for dividend withholding tax), while Hargreaves Lansdown and Morgan Stanley have no subclass and construct `BrokerTransaction` directly.

`BaseParser` stays non-generic, so the type parameter never escapes into the registry: `broker_registry.py` and `main.py` are unchanged. Because `list` is invariant, `load_from_args` widens explicitly in both bases; that is the only place production crosses the boundary, since the registry consumes `load_from_args`.

Test changes follow from the same invariance, in helpers that reach past `load_from_args` to `load_from_file`/`read_transactions`. `on()` now takes a covariant `Sequence`, one call widens for the calculator, and a Vanguard test drops an `isinstance` narrowing that the parser's own type now makes redundant.

### DIFF
--- a/cgt_calc/parsers/base_parsers.py
+++ b/cgt_calc/parsers/base_parsers.py
@@ -41,7 +41,7 @@ class BaseParser(ABC):
         """Load broker data from parsed arguments."""
 
 
-class BaseSingleFileParser(BaseParser):
+class BaseSingleFileParser[T: BrokerTransaction](BaseParser):
     """Parser for single transaction file."""
 
     arg_name: str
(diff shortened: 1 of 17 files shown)


vmartinv (commented):
💯

PR #1038, opened 2026-08-28
New tests and clearer error for schwab-equity-award-json

thibwk:
This is a follow-up to https://github.com/cgt-calc/capital-gains-calculator/pull/662 , where @vmartinv requested more tests.

I took the opportunity to make a few errors clearer while at it.


thibwk:
I'm at a loss as to why that test is failing on Windows only... I tried a few things but no luck. I **think** I'm now doing it like the other parsers, but still no good.

If someone has an idea please let me know...

thibwk:
Found it - I needed `encoding="utf-8"` on the `subprocess.run()`. Now passing.

KapJI:
Thanks! I pushed a commit on top rather than sending it back for another review round.

What it adds:

- Lot validation now runs through the same helper for every sale, including the NVDA split-normalization path. Missing lots, zero or negative prices, and negative share counts are rejected.
- Zero or negative `VestFairMarketValue` values are rejected.
- NaN and Infinity are rejected in parsed numeric fields.
- A missing `SalePrice` now reports the missing field instead of reporting differing prices.
- Cash amounts use the standard amount helper, so `AmountSortValue` works when the display field is absent.

PR #1155, opened 2026-09-19
Order a same-day round trip's buy before its own sell

frederiknolte:
## Problem

`post_process_transactions` sorts every sell on a date ahead of every buy, so the day's disposals fund its purchases (`test_buy_before_same_day_sell_does_not_go_negative` covers that).

A security bought *and* sold on the same day breaks under that rule: its pool is still empty when the disposal is read, and ingestion refuses it with

```
ERROR: Tried to sell not owned symbol, reversed order? for the following transaction:
  2025-02-26 Sell 50 JD, price 34.231791044 GBP, ...
```

The export that hit this has, on one date, `Buy 50 JD`, `Sell 50 JD`, `Buy 50 JD`.

Promoting the round trip's buy ahead of its own sell fixes that one shape and breaks its mirror: a holding carried in from an earlier day, sold on this one to fund buying it straight back, would have the purchase read first from a zero cash balance.

## Approach

Neither side needs promoting. Trades in the same security on the same day are read in the order the statement lists them, and the group sorts under the rank the first of those trades earns — so a sell listed first still leads the day, and an ordinary buy still trails. Rows that are not buys or sells, including the withheld tax that IBKR lists before its dividend, keep the rule they had.

Where the sale falls within the day does not change the gain — TCGA 1992 s105 matches the day's acquisitions and disposals with each other whatever order they arrive in — so this only settles the order ingestion reads them in.

## Changes

- `_trade_ranks` maps each `(date, symbol)` that is traded to the rank of the first of its trades.
- `_by_date_and_action` hands every trade in that group that same rank, and the sort being stable leaves them in the listed order.
- Tests for both orders: `Buy, Sell, Buy` in one security, and a holding sold from a zero balance to buy it back. The first fails on `main`, and the existing same-day ordering and withholding tests still pass.

## Checks

`ruff`, `ruff format`, `mypy`, `ty`, Harper and the full `pytest` suite (2021 passed) all pass locally.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


vmartinv (changes requested):
I think this won’t work in this case:
Case 1: User already owns 5 units of FOO with zero cash balance. On the same day they sell the 5 units and then buy 5 with the proceeds.

Let’s make sure to add this test case. The presented solution is kind of patchy. My suggestion would be to keep it simple: maintain the current order but make it stable within transactions of the same symbol in the same day.

frederiknolte:
You're right — that case breaks it. Owning the shares already and selling to fund the buy-back is exactly the order the round-trip rule would invert, and it isn't a shape the patch could special-case its way out of.

Switched to your suggestion. `_trade_ranks` gives every BUY/SELL of one `(date, symbol)` the rank the first of them earns, and the sort is stable, so those rows keep the order the statement lists them in while everything else is unchanged: other sells still lead the day, ordinary buys and withheld tax still trail. Both orders are covered now — `Buy, Sell, Buy` in one security, and a holding sold from a zero cash balance to buy it straight back.

One thing I noticed while testing: since #1104 judges the balance at each day's close, the action-based promotion no longer seems load-bearing at all — the whole suite (2021 tests) passes with the sort reduced to a plain stable sort on date. I've left it in place here rather than widen this PR, but it may be worth removing separately.

